PR #3, opened 2022-04-21
Upgrade yarn 3

tiziodcaio:
Removed pnp, so the compatibility should be the same as yarn v1


vshcherb:
Will check it next wee.

vshcherb:
Tested and works

PR #6, opened 2022-05-08
Fix link to navigation entry

WPettersson:
The navigation entry incorrectly pointed to the plugins page.


vshcherb:
Should be available for test https://maptile.osmand.net/

PR #18, opened 2022-06-20
add favorites

alisa911:


map/src/components/OsmAndMap.js
```diff
@@ -69,9 +58,43 @@ async function addTrackToMap(ctx, file, map) {
...
+        if (response.ok) {
+            trackData = await response.text();
+        } else {
+            trackData = '<gpx version="1.1" />'
+        }
+    }
+    return trackData;
+}
+
+async function addFavoritesToMap(ctx, file, map) {
```
vshcherb, line 83: Likely it should be refactored and Tracks + Favorites should be in separate layers https://github.com/osmandapp/web/tree/e280e0d8997736a08d8d0c93b9efdd1db8c4a83a/map/src/components/layers

map/src/context/AppContext.js
```diff
@@ -356,7 +358,10 @@ export const AppContextProvider = (props) => {
         routeMode, setRouteMode,
         weatherPoint, setWeatherPoint,
         routeTrackFile, setRouteTrackFile,
-        searchCtx, setSearchCtx
+        searchCtx, setSearchCtx,
+        favoriteFile, setFavoriteFile,
```
vshcherb, line 362: looks like too many state variables for such top level and too much of internal structure exposed

map/public/images/map-icons-png
```diff
@@ -0,0 +1 @@
+../../../../resources/rendering_styles/style-icons/map-icons-png/drawable-xxhdpi
```
vshcherb, line 1: we should try to use svg icons

PR #24, opened 2022-07-23
clarify how to use the `repo` utility in combination with the `readonly.xml`

TheSnoozer:


TheSnoozer:
Thanks for getting this integrated!

PR #30, opened 2022-08-15
Local track 2.1

alisa911:


map/src/util/Utils.js
```diff
@@ -111,24 +115,38 @@ async function uploadFile(gpxFiles, setGpxFiles, ctx, gpxLayer, file) {
...
     let file = Object.assign('', currentFile);
-    file.name = file.name.replace(/.gpx/, '');
-    if (file.name.includes('/')) {
-        return file.name.split('/')[1]
-    } else if (file.local && file.name.includes(':')) {
-        return file.name.split(':')[1]
+    return prepareName(file.name, file.local);
+}
+
+function prepareName(name, local) {
```
vshcherb, line 121: Bad function for global utility probably should not be here. 
Global utility should have only common algorithms no business logic like name formatting by unknown rules.
Good Function:
- stripExtension(fileName);

Bad Function
- formatGpxName(fileName);

map/src/contextmenu/LocalTrackTabList.js
```diff
@@ -0,0 +1,29 @@
+import LocalInfoTab from "./components/tabs/LocalInfoTab";
```
vshcherb, line 1: Should be definitely combined with TrackTabList in future there is no reason that ElevationTab or SpeedTab won't be accessible for local tracks, so it's non future proof.

PR #50, opened 2022-11-28
Fix a recurring typo

guntbert:
parkig --> parking


vshcherb:
I guess file names also should be changed

guntbert:
@vshcherb what filenames?

vshcherb:
```
'built-up-areas/built-up_ParkigPermissive',
```
Looks like a file name not a string

guntbert:
thx for catching my blunder :-/
Do you have any idea, whether those referenced files reside in osmand? If not I'd rather change only the text part for now.

vshcherb:
Nope it's only here used in this repo

guntbert:
@vshcherb I think I have found and renamed the correct files.
I have renamed all of them, not only the ones mentioned in my first commit.

PR #86, opened 2023-05-18
Rename index.md to index.kmz

tammynewbie:
Map


tammynewbie (commented):
Charge more than displayed amount for service that was different than described

PR #114, opened 2023-06-14
Editing tracks without routing, zip localstorage

RZR-UA:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2053

RZR-UA:
gzip is already included in #115

RZR-UA:
- [ ] check possible bugs in case when old (uncompressed) tracks are alredy loaded

- [ ] maybe good idea to change LOCAL_TRACK_KEY to new COMPRESSED_TRACK_KEY

- [ ] another good idea to make compatibility function to turn uncompressed tracks into compressed (let's keep user's existing localTracks safe)

PR #122, opened 2023-06-28
Fix loadTracks() bugs, better errors processing

RZR-UA:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2051

alisa911:

PR #130, opened 2023-07-13
rzr bugs 13jul

RZR-UA:
**Fix TrackMenu Expand: don't depend it on gpxfiles**
(this will show Expand icon always, not based on gpxfiles count, because Tracks always has submenu Local)

**Fix bug with stale context ("layers is undefined")**
(this will fix dangerous bug with outdated context in map.on click handler in Tracks Editor)

**Fix lost query string and #hash on login/logout**
(this will track zoom/geo (hash) and route params (qs) between login/logout navigation)

**Fix bug: don't hide PanelButtons while Track edit**
(this won't hide PanelButtons where right drawer is hidden)

### DIFF
--- a/map/src/map/layers/LocalClientTrackLayer.js
+++ b/map/src/map/layers/LocalClientTrackLayer.js
@@ -67,6 +67,20 @@ export default function LocalClientTrackLayer() {
                 checkUpdateLayers();
             }
         }
+        if (ctx.selectedGpxFile?.layers) {
+            /*
+                map.off('click');
+                map.on('click', clickMap);
+
+                clickMap() uses context vars?
+                Got old context inside event handler?
+
+                1. Use Effect() to monitor context changes.
+                2. Disable previous handler (don't queue handlers).
+                3. Re-setup new event handler with refreshed context.
+             */
+            addClickOnMap(); // refresh ctx
+        }
     }, [ctx.selectedGpxFile]);
 
     useEffect(() => {
@@ -101,7 +115,6 @@ export default function LocalClientTrackLayer() {
 
         setLocalLayers({...localLayers});
     }, [ctx.localTracks, ctx.setLocalTracks]);
-
 
     useEffect(() => {
         if (ctx.createTrack?.closePrev && !_.isEmpty(ctx.createTrack.closePrev.file)) {
@@ -561,7 +574,8 @@ export default function LocalClientTrackLayer() {
         } else if (ctx.selectedGpxFile?.dragPoint === undefined) {
             ctx.trackState.block = false;
             map.getContainer().style.cursor = 'crosshair';
-            map.on("click", clickMap);
+            map.off('click'); // !!!
+            map.on('click', clickMap);
         }
     }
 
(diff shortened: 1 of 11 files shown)


map/src/map/layers/LocalClientTrackLayer.js
```diff
@@ -67,6 +67,20 @@ export default function LocalClientTrackLayer() {
...
+                map.on('click', clickMap);
+
+                clickMap() uses context vars?
+                Got old context inside event handler?
+
+                1. Use Effect() to monitor context changes.
+                2. Disable previous handler (don't queue handlers).
+                3. Re-setup new event handler with refreshed context.
+             */
+            addClickOnMap(); // refresh ctx
```
alisa911, line 80: Wrong place to add click. Need add click only in https://github.com/osmandapp/web/blob/d82e763f18863f1238629acc366ca8669af1d50e/map/src/map/layers/LocalClientTrackLayer.js#L106
We add click when start drawing and stop click when finish drawing.

map/src/map/layers/LocalClientTrackLayer.js
```diff
@@ -67,6 +67,20 @@ export default function LocalClientTrackLayer() {
                 checkUpdateLayers();
             }
         }
+        if (ctx.selectedGpxFile?.layers) {
+            /*
+                map.off('click');
+                map.on('click', clickMap);
+
+                clickMap() uses context vars?
```
alisa911: Delete comments with dialogs

map/src/map/layers/LocalClientTrackLayer.js
```diff
@@ -561,7 +574,8 @@ export default function LocalClientTrackLayer() {
         } else if (ctx.selectedGpxFile?.dragPoint === undefined) {
             ctx.trackState.block = false;
             map.getContainer().style.cursor = 'crosshair';
-            map.on("click", clickMap);
+            map.off('click'); // !!!
```
alisa911, line 575: The method is called addClickOnMap and we only add the click, not remove it

map/src/map/layers/LocalClientTrackLayer.js
alisa911, line 80: Need to fix getTrackWithAnalysis - replace ctx with current ctx.selectedGpxFile (trackRef.current for createNewRouteLine)

alisa911 (changes requested):

map/src/map/layers/LocalClientTrackLayer.js
RZR-UA: resolved by full refactor add/del click (now both are in one place)

map/src/map/layers/LocalClientTrackLayer.js
RZR-UA: renamed to setClickOnMap, because it's called more than once anyway

map/src/map/layers/LocalClientTrackLayer.js
RZR-UA: comment fixed

map/src/map/layers/LocalClientTrackLayer.js
RZR-UA: dealt another way

RZR-UA:
refactored

PR #132, opened 2023-07-14
rzr bugs 14 jul - for linter branch

RZR-UA:
Refactored for the lintered branch.

**Fix TrackMenu Expand: don't depend it on gpxfiles**
(this will show Expand icon always, not based on gpxfiles count, because Tracks always has submenu Local)

**Fix bug with stale context ("layers is undefined")**
(this will fix dangerous bug with outdated context in map.on click handler in Tracks Editor)

**Fix lost query string and #hash on login/logout**
(this will track zoom/geo (hash) and route params (qs) between login/logout navigation)

**Fix bug: don't hide PanelButtons while Track edit**
(this won't hide PanelButtons where right drawer is hidden)

### DIFF
--- a/map/src/context/AccountManager.js
+++ b/map/src/context/AccountManager.js
@@ -1,6 +1,6 @@
 import { apiGet, apiPost } from '../util/HttpApi';
 
-export const LOGIN_LOGOUT_URL = '/map/loginForm#logout';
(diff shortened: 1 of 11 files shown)


alisa911:

RZR-UA:
#2078

PR #140, opened 2023-07-19
fix poi icons, dialog

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/1903

map/src/context/PoiManager.js
```diff
@@ -61,18 +84,28 @@ function getIconNameForPoiType(iconKeyName, typeOsmTag, typeOsmValue, iconName)
     }
 }
 
+function formattingPoiFilter(type, rename) {
+    if (type) {
+        if (rename) {
+            type = poiFilters[`${type}`].length > 1 ? poiFilters[`${type}`][1] : poiFilters[`${type}`][0];
```
RZR-UA, line 90: use **type** instead of ${type}

map/src/context/PoiManager.js
```diff
@@ -61,18 +84,28 @@ function getIconNameForPoiType(iconKeyName, typeOsmTag, typeOsmValue, iconName)
     }
 }
 
+function formattingPoiFilter(type, rename) {
+    if (type) {
```
RZR-UA, line 88: Why not if (type && rename)?

map/src/drawer/components/poi/PoiTypesDialog.jsx
```diff
@@ -100,6 +100,23 @@ export default function PoiTypesDialog({ dialogOpen, setDialogOpen }) {
         return categories.filter((category) => category !== 'user_defined_other' && category !== 'osmwiki');
     }
 
+    function removeUnusedFilters(filters) {
+        filters = filters.filter((f) => f !== 'routes');
```
RZR-UA, line 104: Do you really need to modify function parameter and then return it also? If not, just use **return filters.filter(...)**

map/src/drawer/components/poi/PoiTypesDialog.jsx
```diff
@@ -186,12 +203,12 @@ export default function PoiTypesDialog({ dialogOpen, setDialogOpen }) {
...
                                 />
                             }
-                            label={`${category}`}
+                            label={getLabel(category)}
                         />
                     ))}
                 </div>
                 <Grid container spacing={2}>
-                    {ctx.poiCategory?.filters.map((item, key) => (
+                    {removeUnusedFilters(ctx.poiCategory?.filters).map((item, key) => (
```
RZR-UA, line 211: Need key for Grid (as 1st element under map-cycle)

RZR-UA (changes requested):

RZR-UA (approved):

PR #141, opened 2023-07-19
Fix avoidRoads and maxDist bug in calculateRoute

RZR-UA:


RZR-UA:
branch merged successfully

PR #144, opened 2023-07-20
Backup menu

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/1802

map/src/login/DownloadBackupDialog.jsx
```diff
@@ -68,6 +69,13 @@ export default function DownloadBackupDialog({ openDownloadBackupDialog, setOpen
             let groups = getCategoryGroups();
             groups = prepareFileGroups(groups);
             setCategories(groups);
+            let res = [];
```
RZR-UA, line 72: use const (with const you can modify const [] and {} but can't reassign it to another arr/obj)

map/src/login/DownloadBackupDialog.jsx
```diff
@@ -147,17 +155,22 @@ export default function DownloadBackupDialog({ openDownloadBackupDialog, setOpen
...
         names.forEach((n) => {
             if (userPlaces.includes(n)) {
-                res = addToGeneralGroup('My places', res, n, groups);
-            }
-            if (settings.includes(n)) {
-                res = addToGeneralGroup('Settings', res, n, groups);
-            }
-            if (resources.includes(n) || n.startsWith(`${FILE_TYPE}_`)) {
-                res = addToGeneralGroup('Resources', res, n, groups);
+                addToGeneralGroup([menuType[0]], res, n, groups);
```
RZR-UA, line 167: Very muddled if:

1) why indexes 0 1 3 2 instead of 0 1 2 3 were used?

2) no final else {} (what will happen when all if will got false)

Please think about clarifying this code.

map/src/login/DownloadBackupDialog.jsx
```diff
@@ -204,6 +217,17 @@ export default function DownloadBackupDialog({ openDownloadBackupDialog, setOpen
         }
     }
 
+    function selectAllCategory(type, res, groups) {
```
RZR-UA, line 220: As for me, it's not good idea to modify array/object which is parameter of the function.

This is OK only when it's absolutely clear by code (for example, rename function to selectAllCategoryAndUpdateRes()

In your case, it's better to make new array inside function, push res to it, then push other elements, return it and use result by parent as new array.

RZR-UA (approved):

PR #147, opened 2023-07-24
fix mobile version

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2064

map/src/drawer/components/GeneralPanelButtons.jsx
```diff
@@ -37,11 +43,25 @@ export default function GeneralPanelButtons({ drawerWidth }) {
         });
     };
 
+    function getOrientation() {
+        return height > 500 && width > 0 ? 'vertical' : 'horizontal';
+    }
```
RZR-UA, line 48: Are you sure this is the best method to get orientation?

map/src/infoblock/components/InformationBlock.jsx
```diff
@@ -86,11 +96,18 @@ export default function InformationBlock({ hideContextMenu, drawerWidth }) {
...
+                        setDrawerHeight(50);
+                    }
                 }
             }
         }
     }, [ctx.currentObjectType, ctx.selectedGpxFile, ctx.weatherPoint, ctx.updateContextMenu]);
 
+    useEffect(() => {
+        getWidth();
+    });
```
RZR-UA, line 109: Effect shouldn't run every render.

PR #148, opened 2023-07-24
rzr bugs 24jul

RZR-UA:
**Add distance Total/Segment instead of "FromStart"**

- [x] test: distance for multi-segment gpx (take into account that charts show multi-segments!)

- [x] test: distance for created track with gap/gaps (point context menu = split after/before)

**Fix Total distance bug, improve osrm/line distance**

- [x] test distance (total in header, in info-block, in graphics) for different track route providers and profiles, including native- and srtm-graps - all distance results should be equal +- 0.1km

**Fix showPoints Effect bug, test useMutator wrapper:** 

- [x] test showPoints keep working with Undo/Redo and Add new point in Track editor

**Add extra space to input length for Track EditName:**
 
- [x] test Track EditName with different combinations of lat/non-lat letters, lowercase and uppercase, numbers & special characters

### DIFF
--- a/map/src/infoblock/components/track/GeneralInfo.jsx
+++ b/map/src/infoblock/components/track/GeneralInfo.jsx
@@ -197,6 +197,9 @@ export default function GeneralInfo({ width, setOpenDescDialog }) {
     };
 
     const EditName = () => {
+        const isLetterUpperCase = (l) => l === l.toUpperCase() && l.toLowerCase() !== l.toUpperCase();
+        const nUpperCaseLetters = fileName.split('').filter((c) => isLetterUpperCase(c)).length;
+        const inputLength = fileName.length + nUpperCaseLetters + 3; // add extra space
         return (
             <div style={{ display: 'flex', maxWidth: '400px', flexWrap: 'wrap' }}>
                 {!disableButton && (
@@ -204,7 +207,7 @@ export default function GeneralInfo({ width, setOpenDescDialog }) {
                         <TextareaAutosize
                             style={{
                                 maxWidth: '400px',
-                                width: fileName.length + 'ch',
+                                width: inputLength + 'ch',
                                 resize: 'none',
                                 marginBottom: '5px',
                                 fontSize: '16px',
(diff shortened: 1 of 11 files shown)


RZR-UA:

map/src/infoblock/components/track/GeneralInfo.jsx
```diff
@@ -331,7 +334,7 @@ export default function GeneralInfo({ width, setOpenDescDialog }) {
                         </ListItemIcon>
                         <ListItemText>
                             <Typography sx={{ ml: 1 }} variant="body2" noWrap>
-                                {`Elevation SRTM (min/avg/max): ${elevationSRTM}`}
+                                {`Elevation SRTM: ${elevationSRTM}`}
```
alisa911, line 337: Why did you delete (min/avg/max)?

map/src/util/Utils.js
```diff
@@ -1,5 +1,35 @@
...
+ * Create object copy -> Apply callback() -> Call setter()
+ *
+ * Example:
+ *
+ * const [showPoints, mutateShowPoints] = useMutator({ points: true, wpts: true });
+ *
+ * mutateShowPoints((next) => (next.points = !next.points)) // functional way
+ * mutateShowPoints({ points: true, wpts: true }) // classic way supported
+ */
+export function useMutator(init) {
```
alisa911, line 16: Should to create a folder with name hooks

map/src/infoblock/components/track/GeneralInfo.jsx
RZR-UA: This is temporarily change to fit Ele-SRTM string into 1 line. Let's improve it later when testing mobile version.

map/src/util/Utils.js
RZR-UA: Agreed. Will be done in Ksenia's branch.

EugeneZmeuk:
- [ ] Distance / 2:
   - open  Rec track ( !my track has two segments):
   
[2023-07-16_14-03_Sun.gpx.zip](https://github.com/osmandapp/web/files/12160402/2023-07-16_14-03_Sun.gpx.zip)

image

   - recalculate Elevation SRTM and receive the wrong distance for the graph:
image
   
 - [ ]  not give permission to choose "All segments" (recording part of the chosen track) for Car, Bike and Pedestrian profiles. Because the algorithm can't recalculate old (rec) part of the chosen track. When we try to stop process by Undo we can't work properly with this track.
 
image

    - Distance for Edit

PR #151, opened 2023-07-26
Fix Effect deps for getWith, fix setState warning

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2064

PR #153, opened 2023-07-26
rzr bugs 26jul

RZR-UA:
**Fix auto-Visible tracks, fix "empty-tracks" issues**

- [ ] test: check that any kind of created tracks was appeared in Visible tracks (any: from context menu, by pen-icon, by create-button, by gpx-upload, by click on local tracks, etc)

**Disable edit Description for non-local Tracks**

- [ ] test: track description should be not-editable when you view Cloud tracks (without Edit them) - bug fixed

**Sanitize trackName in Editor, fix connected bugs**

- [ ] test: now you can change track name for just created track even w/o points - crash fixed

- [ ] test: try any bogus/dangerous/multiline text as track name, then save and check results - sanitize added

**Fix never-closing deleteLocalTrack on empty track**

- [ ] test: bar-button Delete should successfully handle just created track w/o points - bug fixed


RZR-UA:

PR #156, opened 2023-07-27
rzr bugs 27jul

RZR-UA:
**Fix bug with lost state after closePrev processed**

- [x] test: when you have several Local tracks, test click between them in Local tracks list, every click should start Edit track (compare behavior with the production)

**Remake LocalTrackItem, drop track.index, fix keys**

- [x] test: check that created track moves to Visible once after created+closed

- [x] test: sorted A-Z Local tracks should work (click/edit should choose right track)

- [x] test: deleted Local tracks should not cause crashes when using Visible tracks list

**Improve fitBounds for empty track, add maxZoom: 17**

- [x] test: create empty track should not do zoom-to-fit

- [x] test: track with 1 point should not do zoom-to-fit

- [x] test: very small track (2-3 segments, 10 meters each) should not zoom-to-fit closer than 17

- [x] test fitBounds (should works) on a few external gpx-tracks


RZR-UA:

PR #157, opened 2023-07-28
Hide some features until develFeatures true #2099

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2099

PR #160, opened 2023-07-29
rzr 28jul: polish main menu, revamp Cloud upload, fix state bugs, fix other bugs

RZR-UA:
# PLEASE USE BRANCH rzr-bugs-30jul FOR TESTS

**Revamp Cloud upload: auto-activate, safety, bugfix**

- [x] test (next branch): multiple Edit <-> Save to Cloud for any track

- [x] test: Saved to Cloud track should be auto-activated

- [x] test (next branch): Overwrite Cloud track with another Local track and check it's refreshed (in Cloud tracks) automatically

- [x] test (next branch): Cloud overwrite-warning should be shown every try (there was a bug - no warning was shown 50/50)

- [x] test: Local track should be deleted after Save to Cloud

- [x] test: Safety: Local track should not be deleted in case of Upload error (to test, load Local track, remove Cookies, then press Save to Cloud and check results)

**Polish menu (check/test):**

- [x] Hide Favorites menu when nothing to show inside

- [x] Show Local+Cloud track count instead of just Cloud

- [x] Open Local tracks (submenu) by default to show Upload/Create

**Minor bugs (check/test):**

- [x] Fix crash loading Visible Cloud track after (!) Logout

### DIFF
--- a/map/src/drawer/components/tracks/CloudTrackItem.jsx
+++ b/map/src/drawer/components/tracks/CloudTrackItem.jsx
@@ -3,7 +3,7 @@ import { Alert, LinearProgress, ListItemText, MenuItem, Switch, Tooltip, Typogra
 import React, { useContext, useState } from 'react';
 import Utils from '../../../util/Utils';
 import TrackInfo from './TrackInfo';
-import TracksManager from '../../../context/TracksManager';
+import TracksManager, { isEmptyTrack } from '../../../context/TracksManager';
 
 export default function CloudTrackItem({ file }) {
     const ctx = useContext(AppContext);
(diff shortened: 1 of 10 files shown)


RZR-UA:

map/src/drawer/components/tracks/CloudTrackItem.jsx
```diff
@@ -31,35 +31,34 @@ export default function CloudTrackItem({ file }) {
...
+            const f = await Utils.getFileData(newGpxFiles[file.name]);
             const gpxfile = new File([f], file.name, {
                 type: 'text/plain',
             });
-            let track = await TracksManager.getTrackData(gpxfile);
+            const track = await TracksManager.getTrackData(gpxfile);
             setProgressVisible(false);
-            if (track) {
-                let type = ctx.OBJECT_TYPE_CLOUD_TRACK;
+            if (track && (track.points?.length > 0 || track.wpts?.length > 0 || track.tracks?.length > 0)) {
```
vshcherb, line 52: this is a bad pattern, if someone will add new field that method won't work and nobody is going to fix. It's unclear by the code why empty points  track is not accepted which is not bad probably, and anyway such checks should be part of track object itself to make it more encapsulated like isEmpty method however I would question if it's really needed.

alisa911 (approved):

map/src/drawer/components/tracks/CloudTrackItem.jsx
RZR-UA: TracksManager.isEmptyTrack was added

PR #161, opened 2023-07-31
fix param undefined, fix show fav list

alisa911:
- [ ] fix create new favorite without groups
- [ ] fix create new group and save to it
- [ ] fix open favorites list


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2133

EugeneZmeuk:
|31.07.2023|
|-----------|

1.

| Can't choose "Select group" full icon (only first three folders)|
|-------------|
|image|

PR #164, opened 2023-08-01
fix looping after selecting a group without favorites after deleting

alisa911:


map/src/drawer/components/favorite/FavoriteGroup.jsx
```diff
@@ -122,7 +124,7 @@ export default function FavoriteGroup({ index, group, enableGroups, setEnableGro
                 };
                 markerList.push(marker);
             });
-        } else if (favoritesPointsOpen && markers.length === 0) {
+        } else if (favoritesPointsOpen && markers.length === 0 && FavoritesManager.getGroupSize(group)) {
```
RZR-UA: Did you mean getGroupSize(...) > 0 ?

map/src/context/FavoritesManager.js
```diff
@@ -188,6 +188,10 @@ function createDefaultWptGroup(wptGroup) {
     }
 }
 
+function getGroupSize(group) {
+    return group && group?.pointsGroups[group.name === DEFAULT_GROUP_NAME ? '' : group.name]?.points.length;
+}
+
```
RZR-UA: get...Size() implies number as return type

please add default "return 0"

map/src/drawer/components/favorite/FavoriteGroup.jsx
```diff
@@ -79,36 +79,38 @@ export default function FavoriteGroup({ index, group, enableGroups, setEnableGro
...
-                newSelectedGpxFile.file = Object.assign({}, newFavoriteFiles[group.name]);
-                newSelectedGpxFile.nameGroup = group.name;
-                if (!ctx.selectedGpxFile) {
-                    ctx.selectedGpxFile = {};
-                }
-                if (ctx.selectedGpxFile.markerCurrent) {
-                    newSelectedGpxFile.markerPrev = ctx.selectedGpxFile.markerCurrent;
-                }
-                ctx.setSelectedGpxFile(newSelectedGpxFile);
+        if (FavoritesManager.getGroupSize(group) > 0) {
```
RZR-UA, line 82: Please avoid using OBJECT.something() call because this way is not controlled by the linter.

For external usage it's better to import exact exported function instead of using exported default object.

For local usage (in this module) it's better to call direct function.

PR #166, opened 2023-08-01
Fix Popper's anchorEl and createPointContextMenu()

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2137

alisa911 (approved):

PR #167, opened 2023-08-02
Adapt PointContextMenu to mobile touch-events, fix mobile events

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2137

map/src/map/EditableMarker.js
```diff
@@ -63,12 +64,35 @@ export default class EditableMarker {
...
+                if (e.type === 'touchend') {
+                    e.preventDefault();
+                    if (e.target?.title !== 'poly') {
+                        e.target?.dispatchEvent(new Event('contextmenu', e));
+                    }
+                }
+            }
+        };
+        document.addEventListener('touchstart', markerEventHandler, { passive: false });
+        document.addEventListener('touchend', markerEventHandler, { passive: false });
```
RZR-UA, line 87: We need only 1 eventHandler added to document.

Please move addEventListener to one-time useEffect(..., []) into component which permanently exists as only one instance. For example, into LocalClientTrackLayer.

PR #175, opened 2023-08-07
fix open empty info block

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2140

RZR-UA (approved):

PR #180, opened 2023-08-09
Add beep voice pack description

mikehgentry:
I think I made a mess of committing the PR, but the actual edit hopefully makes sense!

The only bit I'm unsure of is the Translate markdown, which I hope displays the buttons to press to get to recorded voice packs, but doesn't show up in preview.

Cheers


mikehgentry:
The only other thing I can think to do is to edit the wording of the overview to say "Voice prompts, cycle computer style beeps, and text notifications are features..." or something of that nature.

PR #183, opened 2023-08-10
Add yarn start:fallback, use if Test API is broken

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2153

PR #187, opened 2023-08-16
Refactor TrackRoutingCache

RZR-UA:
- [ ] test: check that analytics came after every routing/undo/redo

- [ ] test: create track with 30-40 segments, use undo/redo, use different profiles

- [ ] test: re-route all segments several times, using different profiles, check cache speed

- [ ] test: reload page, repeat re-routing all segments, try undo/redo in-process and after routing

- [ ] test: reload page, repeat re-routing all segments, try re-route before current re-route is done

- [ ] test: try different combinations with drag-drop existing points (only existing - don't drag new point from line)

- [ ] test: try different usage of point context menu, including split/join segments and re-routing after some segments were splitted

- [ ] test: finally, draw attention to the analytics chart after undo/redo, when using different routing types (OSRM = yellow, OsmAnd = blue)


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2166

map/src/context/TracksRoutingCache.js
```diff
@@ -1,78 +1,201 @@
...
+
+    if (updated > 0) {
+        requestAnalytics({ ctx, track, debouncerTimer });
+        saveChanges(null, null, null, track); // mutate track with more data and call setSelectedGpxFile({...})
+    }
+
+    return updated > 0;
+}
+
+export function debouncer(f, timerRef, ms) {
```
alisa911, line 92: code duplication

map/src/util/Utils.js
```diff
@@ -130,6 +155,7 @@ export function prepareFileName(filename) {
         return new TextDecoder().decode(truncated);
     };
 
+    const newlineRe = /\n/g;
```
alisa911, line 158: Add comments about all regex (what does he do and why)

RZR-UA: details added

map/src/context/TracksRoutingCache.js
RZR-UA: Not a duplication. export function debouncer was moved from LocalClientTrackLayer to Cache module and are used from both of them.

EugeneZmeuk:
- [ ] new point 

|adding new point during recalculation route for the new profile|
|------|
|image|

RZR-UA:
adding new point during recalculation route for the new profile

this is old bug, not related to cache and/or cache-refactor

please open new task

PR #194, opened 2023-08-18
Improve points[].geometry[] distance recalculation, fix local-cloud and gpx duplicated geometry #2179

RZR-UA:
- [ ] test: try to repeat "charts-broken-segments" bugs (the bug activated always after uploading to Cloud or re-using Downloaded GPX file)

- [ ] test: create track, do 10 cycles: Local-Cloud...Local-Cloud, every time use download gpx, finally all downloaded gpx should have same size


RZR-UA:

PR #195, opened 2023-08-19
Fixhas geo

alisa911:


alisa911:

map/src/context/TracksManager.js
```diff
@@ -320,26 +320,29 @@ function prepareTrack(track, localName = null, originalName = null) {
 }
 
 function hasGeo(track) {
+    let hasGeo;
```
RZR-UA, line 323: forgotten default value: let hasGeo = false;

Also, need to rewrite whole function to avoid forEach().
Just use for() and return true if geo exists
(in this case we can drop hasGeo flag)

RZR-UA (changes requested):

PR #196, opened 2023-08-19
Improve addDistance() recalculation (segments/total), fix little distance-related bugs, drop outdated code

RZR-UA:
- [ ] test: try to repeat "charts-broken-segments" bugs (the bug was activated always after uploading to Cloud or re-using Downloaded GPX file)

- [ ] test: create track, do 5 cycles: Local-Cloud...Local-Cloud, every time use download gpx, finally all downloaded gpx should have same size

- [ ] test: header stats for Selected tracks should work both for Local and Cloud selected tracks


RZR-UA:

PR #204, opened 2023-08-25
GPX-segments (after routed-segments added): protect against bulk re-routing, allow drag-drop new point, fix NaN in track points distance

RZR-UA:
- [x] open small sized real gpx track, TEST: re-route all segments (should be no actual re-route and no crash), try drag-drop points (first, last, middle), check that charts are refreshed

- [x] add a few Line-profile segments, repeat all tests (should be the same behavior and results)

- [x] add a few non-Line profile segments (a confirmation dialog should appear), try re-route all segments (original gpx segment should have no changes but routed segments should be changed)

- [x] test drag-drop: routed segments, edges between gpx- and routed-segments, try drag new point from gpx-segments, then drag new points from newly generated gpx-segments, etc

- [x] pre-finally, try again re-route all segments (routed should be changed, but **all** gpx-segments should not be re-routed)

- [x] finally, save result (mix of gpx-segments and routed-segments) using the Cloud and then try to Show/Navigate/Plan route (edit) in the App

### DIFF
--- a/map/src/context/TracksManager.js
+++ b/map/src/context/TracksManager.js
@@ -658,35 +658,6 @@ function decodeRouteMode({ routeMode, params }) {
     return draft;
 }
 
-// async function updateRoute(points) {
-//     let result;
-//     if (points?.length > 0) {
-//         result = await apiGet(`${process.env.REACT_APP_GPX_API}/routing/get-route`, {
-//             method: 'POST',
-//             data: points,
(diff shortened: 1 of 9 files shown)


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2139

alisa911 (approved):

PR #205, opened 2023-08-25
add link for delete acc - /map/delete-account

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2162

RZR-UA (approved):
message text changed, cancel delete-account behavior fixed soon

PR #208, opened 2023-08-29
Refactor track list sorting, improve Clear menu for Visible/Local tracks, improve Tooltip for tracks, refactor/speedup Cloud track list rendering

RZR-UA:
# Track menu sorting:

**Apply all tests on Local and Cloud folders.**

- [x] test: check A-Z and Z-A sorting
- [x] test: check Time/reverse sorting
- [x] test: check Distance/reverse sorting
- [x] test: default sorting should be "Sort by time"
- [x] test: sorting icons should hide when only 1 track in folder 
- [x] test: sorted list should be refreshed on Delete track (check both Local/Cloud delete)
- [x] test: several zero-distance tracks (for example, wpts-only) should be additionally sorted A-Z
- [x] test: try create 2-3 local tracks, edit them sequentally, check that the most freshly edited track always moves to the top
- [x] test: remember Local sorting (screenshot), then upload these 2-3 tracks into new Cloud folder and compare sorting in the Cloud

# Misc sorting:

- [x] test: Cloud tracks folders names sorted A-Z
- [x] test: Visible Local tracks names sorted A-Z
- [x] test: Visible Cloud tracks names sorted A-Z

# UI:

- [x] test: Confirmation dialog on Clear Local tracks (submenu)
- [x] test: Real clearing all Visible tracks using "Clear" submenu
- [x] test: Local tracks Tooltips (mouseover) now includes track analytics
- [x] test: Track Tooltips is now located at right side (desktop) and bottom side (mobile)

# Optimization:

**Open Test and Prod, log in both sites using your account.**

- [x] compare speed: initial full reload
- [x] compare speed: hide/show Weather menu
- [x] compare speed: full hide/show main menu
- [x] compare speed: hide/show big Cloud folders
- [x] compare speed: load/close several Cloud tracks
- [x] compare speed: switching between different not-loaded Cloud tracks
- [x] compare speed: switching between different already-loaded Cloud tracks

PR #209, opened 2023-08-30
Fix 0-points tracks, 1-points tracks, improve Visible tracks removal, improve visible Cloud track overwriting #2168

alisa911:
- [x] test: check empty track (try to create empty track in App and upload it to Cloud from App)
- [x] test: check track with 1 route point and 0 waypoints (Create, Move Local-Cloud-Local, Download/Re-upload GPX)
- [x] test: check track with 1 route point and >0 waypoints (Create, Move Local-Cloud-Local, Download/Re-upload GPX)

- [x] test: Visible Cloud track should disappear from Visible menu after removal this Cloud track by Delete button

- [x] test: Set some Cloud track as Visible, then overwrite Local track over that Cloud track, as result you should have no "zombie-layers", just fresh Cloud track

### DIFF
--- a/map/src/map/layers/CloudTrackLayer.js
+++ b/map/src/map/layers/CloudTrackLayer.js
@@ -5,16 +5,22 @@ import TrackLayerProvider from '../TrackLayerProvider';
 import TracksManager from '../../context/TracksManager';
 import { useMutator } from '../../util/Utils';
 
-function addTrackToMap({ ctx, file, map, fit = true } = {}) {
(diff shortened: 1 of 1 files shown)


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2168

RZR-UA (changes requested):
Code is ok, but should be merged after #208 including conflicts resolving.

RZR-UA (changes requested):
under construction

PR #211, opened 2023-08-31
Improve Login dialog: allow browser auto-login, fix annoying login field reset, allow login by Enter key #2175

RZR-UA:
- [x] test: register new user (use new email, enter code, do login, etc)

- [x] test: do tests with Chromium, Safari, Edge (Firefox is already tested)

- [x] test: try login in by pressing Enter key on Login and/or Password fields

- [x] test: try logout/login several times when NO saved auto-login in Browser

- [x] test: try logout/login several times using ONE saved auto-login in Browser

- [x] test: try logout/login several times using SEVERAL saved auto-logins in Browser

- [x] test: after logout, try login using other E-mail (try reset auto-filled Login field to enter another E-mail)


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2175

PR #216, opened 2023-09-08
add road attributes graph

alisa911:


map/src/infoblock/components/graph/RoadAttributesGraphProvider.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+export default function RoadAttributesGraphProvider({ width }) {
+    const ctx = useContext(AppContext);
+
+    const [roadPoints, setRoadPoints] = useState(null);
+
+    useEffect(() => {
+        if (ctx.selectedGpxFile && ctx.currentObjectType === ctx.OBJECT_TYPE_CLOUD_TRACK) {
+            let points = _.cloneDeep(getTrackPoints(ctx.selectedGpxFile));
+            if (points[0].segment) {
```
RZR-UA, line 15: add undefined check

map/src/infoblock/components/graph/RoadAttributesGraphProvider.jsx
```diff
@@ -0,0 +1,134 @@
...
+                setRoadPoints(points);
```
RZR-UA, line 16: please think about calling setRoadPoints only if they were really changed (apply and test on GpxGraph too); good idea to add console.log at begin of useMemo's function to check how often it was called

map/src/infoblock/components/graph/RoadAttributesGraphProvider.jsx
```diff
@@ -0,0 +1,134 @@
...
+    }, [roadPoints]);
+
+    function generatePastelColor() {
+        let R = Math.floor(Math.random() * 127 + 127);
+        let G = Math.floor(Math.random() * 127 + 127);
+        let B = Math.floor(Math.random() * 127 + 127);
+
+        let rgb = (R << 16) + (G << 8) + B;
+        return `#${rgb.toString(16)}`;
+    }
```
RZR-UA, line 59: Colors for track X should be the same after the track was reloaded (or switched off/on or just clicked in menu). Please think about adding custom random() function where you can use init random seed value (JS Math.random not supports random seed).

map/src/infoblock/components/graph/RoadAttributesGraphProvider.jsx
```diff
@@ -113,9 +116,9 @@ export default function RoadAttributesGraphProvider({ width }) {
         if (colors[label]) {
             newColor = colors[label];
         } else {
-            newColor = generatePastelColor();
+            newColor = generatePastelColor(Object.keys(colors).length);
             while (Object.values(colors).includes(newColor)) {
-                newColor = generatePastelColor();
+                newColor = generatePastelColor(Object.keys(colors).length);
             }
```
RZR-UA: If this while {} happens, the browser freeze, because (Object.keys(color).length) never changes here. Please recheck it and remove while {} block is it is true.

map/src/store/geoRouter/legacy/updateRouteBetweenPoints.js
```diff
@@ -70,8 +70,18 @@ async function updateRouteBetweenPointsOSRM({ start, end, geoProfile, ctx }) {
...
+                apiCache: true,
+                params: {
+                    routeMode: geoProfile.profile,
+                },
+                headers: {
+                    'Content-Type': 'application/json',
+                },
+            });
+
+            return approximateResult ? approximateResult.data.points : points;
```
RZR-UA, line 84: Please add additional final check: 

&& approximateResult?.data?.points?.length >= 2

map/src/infoblock/components/graph/RoadAttributesGraph.jsx
```diff
@@ -12,6 +12,7 @@ export default function RoadAttributesGraph({ name, data, width }) {
         indexAxis: 'y',
         responsive: true,
         maintainAspectRatio: false,
+        animation: false,
```
RZR-UA, line 15: Possibly, good idea to disable animation on main GpxGraps too...

PR #231, opened 2023-09-19
add opt chromium

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2220

RZR-UA:
soved without this changes

PR #238, opened 2023-09-22
New geoObject extends geoRouter class: store navigation data, remove points/options from context, and prepare for versatile data conversion #2078

RZR-UA:


RZR-UA:

alisa911 (approved):

PR #253, opened 2023-10-09
Fix main menu /map absolute url to relative

RZR-UA:


RZR-UA:
sorry wrong approach to fix, closing

PR #263, opened 2023-10-15
Replace old bash/misc "yarn generate" commands with pretty new "yarn generate-resources" written as NodeJS script, add "yarn build" dep on generate-resources

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2088

PR #289, opened 2023-12-05
Allow type=osmand routing for all users (but still not as default)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18512

PR #303, opened 2023-12-25
fix removing track tags

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2270

RZR-UA (approved):

PR #306, opened 2023-12-29
[HH-WEB] Finalize route options in Navigation menu

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2332

PR #311, opened 2024-01-08
Generate OsmAnd Docs Map into docmap.json Issue #2341

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2341

main/scripts/docs-map-generator.js
```diff
@@ -0,0 +1,70 @@
+const fs = require("fs-extra");
+
+module.exports = OsmAndDocMapGenerator;
+
+const PATH = "./docs/";
+const URL = "https://osmand.net/docs/";
```
vshcherb, line 6: Chicken & egg problem, we need a running website to get a structure...

vshcherb: CC @alex-osm  - I'm not sure this is the right approach.

alisa911 (approved):

main/scripts/docs-map-generator.js
alex-osm: Agree, why do we need osmand.net hardcoded here?

PR #314, opened 2024-01-15
[WEB-HH] Fix URL hidden route params such as `hhonly`

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2332

PR #315, opened 2024-01-16
Fix 404 error at /premium/device-registration (email url)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2380

PR #322, opened 2024-01-22
Allow OS flags (android|ios):(true|false) for help-structure-generator

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2341

PR #331, opened 2024-02-12
Crash after deleting last track in the folder - Web

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2398

RZR-UA (approved):

Issue #351, opened 2024-03-10
Bug in poi categories on the site


Nine-Tailed:
### Description

Starting with the "routes" category, there has been a shift between icons and category names.

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/12549744/1cd6bde8-dc21-4125-a08f-e3dbbb3ac21f"  width="300" height="auto"/> </th>
  </tr>
</table>

### Steps to reproduce

Open the list of categories on the site.

### Actual result

The current state is shown in the screenshot above.

### Expected result

The category name must match the icon.

### Your Environment (required)

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version:
Android/iOS version:
Device model:
Crash-Logs: ?
```

vshcherb:
Already available

Nine-Tailed:
Thanks.

PR #359, opened 2024-03-14
Fix poi filters

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/351

map/src/frame/components/poi/PoiTypesDialog.jsx
```diff
@@ -37,6 +37,7 @@ export default function PoiTypesDialog({ dialogOpen, setDialogOpen, width }) {
 
     const classes = useStyles();
     const styles = drawerStyles();
+    const activePoiFilters = removeUnusedFilters(ctx.poiCategory?.filters);
```
RZR-UA, line 40: Will crash (inside function and later on activePoiFilters.map) if no ctx.poiCategory.filters exist (undefined).

RZR-UA (approved):

PR #372, opened 2024-04-01
Cloud Change History (Trash) – Web

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2474

map/src/manager/SettingsManager.js
```diff
@@ -0,0 +1,213 @@
...
+    return templateString.replace(/%\d\$s/g, function (match) {
+        let index = parseInt(match.replace('%', '').replace('$s', ''), 10) - 1;
+        return replacements[index];
+    });
+}
+
+export function isFileRestrictedForDownload(file) {
+    return (
+        file.type.toLowerCase() === 'file' &&
+        (file.name.endsWith('.obf') || file.name.endsWith('.sqlitedb') || file.name.endsWith('.tif'))
```
RZR-UA, line 211: Please add toLowerCase() before endsWith()

map/src/menu/MainMenu.js
```diff
@@ -164,6 +175,7 @@ export default function MainMenu({
     //open main menu if currentObjectType was changed
     useEffect(() => {
         if (ctx.currentObjectType) {
+            closeCloudSettings();
```
RZR-UA, line 178: Probably it is not so good idea to fetch list-files API every time when currentObjectType changed.

Please change approach and avoid requesting `list-files` every time.

For example, you can request list-files only when you have registed changes in the files (delete, upload, empty trash, restore, etc).

PS. Additionally, list-files requires login, but guest users should have ability to use "public accessible" features without login.

map/src/manager/SettingsManager.js
```diff
@@ -0,0 +1,213 @@
...
+        if (f.type === 'month') {
+            const nextFileIndex = changes.slice(index + 1).findIndex((f) => f.type === 'file');
+            return nextFileIndex !== -1;
+        }
+        return true;
+    });
+    return changes;
+}
+
+export function formatString(templateString, replacements) {
```
RZR-UA, line 201: Please add comment with examples of input/output to clarify what the magic format it is.

map/src/menu/settings/CloudChanges.jsx
```diff
@@ -0,0 +1,169 @@
...
+
+        const handleClick = useCallback(
+            (e) => {
+                setOpenActions(true);
+                e.stopPropagation();
+            },
+            [setOpenActions]
+        );
+
+        function getStatus(file) {
```
RZR-UA, line 63: Please add some kind of default status finally after if(s).

map/src/menu/settings/CloudChanges.jsx
```diff
@@ -0,0 +1,169 @@
...
+    useEffect(() => {
+        setChanges(files);
+    }, [files]);
+
+    function closeChanges() {
+        setOpenCloudSettings(false);
+        refreshGlobalFiles({ ctx }).then();
+    }
+
+    const ChangesItem = React.memo(({ item }) => {
```
RZR-UA, line 47: 2 cases of using React.memo. Possibly better to switch to useMemo() with non-object dependency. Let's discuss in PM.

map/src/menu/settings/DeleteFileVersionDialog.jsx
```diff
@@ -0,0 +1,52 @@
...
+        setDialogOpen(!dialogOpen);
+    };
+
+    async function deleteF() {
+        if (deleteVersion) {
+            deleteFileVersion(file, ctx, changes, setChanges).then();
+        } else if (deleteAllVersions) {
+            deleteFileAllVersions({ file, changes, setChanges, isTrash: true }).then();
+        } else {
+            deleteFile(file, ctx, changes, setChanges).then();
```
RZR-UA, line 33: 1 function = 2 kind of parameters pass for child calls...

Please change all the deleteFile, deleteFileVersion to ({...})

map/src/menu/settings/DeleteFileVersionDialog.jsx
```diff
@@ -0,0 +1,52 @@
...
+    deleteVersion = false,
+    deleteAllVersions = false,
+}) {
+    const ctx = useContext(AppContext);
+    const { t } = useTranslation();
+    const toggleShowDialog = () => {
+        setDialogOpen(!dialogOpen);
+    };
+
+    async function deleteF() {
```
RZR-UA, line 27: When a function is actually "async", it is prettier to use `await child()` instead of child().then()

map/src/menu/settings/DeleteFileVersionDialog.jsx
```diff
@@ -0,0 +1,52 @@
...
+        <Dialog open={true} onClose={toggleShowDialog}>
+            <DialogTitle>Delete</DialogTitle>
+            <DialogContent>
+                <DialogContentText>
+                    {formatString(t('delete_trash_item_confirmation_desc'), [file.name])}
+                </DialogContentText>
+            </DialogContent>
+            <DialogActions>
+                <Button onClick={toggleShowDialog}>Cancel</Button>
+                <Button onClick={() => deleteF()}>Delete</Button>
```
RZR-UA, line 48: Please simplify {...} -> {deleteF}

Why Cancel and Delete have different approach?

RZR-UA (changes requested):
Finnally:

1) please verify that site works good for anonymous users (before login) - there should be no list-files and other "authorized" API requests

2) please avoid redundant list-files requests because they can be slow (both on server and browser); try using EZ account to test this branch

map/src/menu/MainMenu.js
alisa911: I looked into it, which is why I implemented it as it is now. 

The regular 'list-files' used in tracks only retrieves tracks and favorites, and only the latest versions if not deleted.

In this case, however, all versions of all types of files are retrieved, and there are many more of them. Therefore, I implemented lazy loading, and that’s why after any actions, I do not make a redundant heavy request for all versions, but work with what is already there, updating the existing list. And that’s why the update of the old 'list-files' is made upon exiting the cloud menu. Without this update, it would not be possible.

alisa911:
> Finnally:
> 
> 1. please verify that site works good for anonymous users (before login) - there should be no list-files and other "authorized" API requests
> 2. please avoid redundant list-files requests because they can be slow (both on server and browser); try using EZ account to test this branch

All the added functionality is only for logged-in users, as it involves cloud operations. Without logging in, a user won’t even see these options. So, everything is set up correctly in this regard.

map/src/menu/settings/CloudChanges.jsx
alisa911: useMemo is a hook for memoizing values within a component to avoid expensive recalculations on re-renders with unchanged dependencies. React.memo is a higher-order component that prevents re-rendering if a component's props haven't changed, optimizing performance. Thus, I chose React.memo to optimize rendering behavior based on prop changes, not for memoizing internal calculations.

RZR-UA: Ok. We've never used React.memo for this purpose before.

Using useMemo was enough to create memoized child components.

Anyway React.memo might be OK, but please test how React.memo works in your case:

- item is not an object so possibly you can't use it directly as dependency (for example, you definitely can not use this item in useEffect as a ..., [dep])

- try logging and monitor how often both React.memo() components really rendered (use 1st console.log before React.memo and 2nd console.log just inside component's { })

map/src/menu/MainMenu.js
RZR-UA: Let's test the final version. For the moment, I can not see list-files after any currentObjectType changed, anymore.

RZR-UA (approved):

PR #373, opened 2024-04-02
Get back 'end' instead of 'finish' (compatibility)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19422

PR #386, opened 2024-04-19
add poi icons to tags

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2320

RZR-UA (approved):

PR #388, opened 2024-04-20
Fix GPX export name not updating

### LINKED ISSUE #387: Track name is not updated in GPX export xml
For example, create a track which initially has the name of `Sat Apr 20 2024`, update the name to `Test`, the name in the GPX file remains `Sat Apr 20 2024`:
```xml
<metadata>
  <name>Sat Apr 20 2024</name>
</metadata>
  ```

I have a fix working locally, which I'll submit a PR with. This is my first time working with OsmAnd though so the fix may not be correct, hence why I created this issue separately.

### PR DESCRIPTION
**Pull request by FozzieHi**
Fixes #387

This updates `prepareTrackData` to also set `file.metaData.name` to `file.name` if there is a difference between the two strings. This correctly updates the GPX export XML:
```xml
<metadata>
  <name>Test</name>
</metadata>
```

I also made a quick change to `file.tracks`, but I'm happy to change this back if it doesn't fit in with the project's code style.

### DIFF
--- a/map/src/manager/track/TracksManager.js
+++ b/map/src/manager/track/TracksManager.js
@@ -450,16 +450,13 @@ export async function getGpxFileFromTrackData(file) {
 
 function prepareTrackData({ file, getAnalysis = false }) {
     // add updated points to track
-    if (file.tracks && file.tracks[0] && validateRoutePoints(file.points)) {
+    if (file.tracks?.[0] && validateRoutePoints(file.points)) {
         file.tracks[0].points = file.points;
     }
 
     // add metaData name if it isn't exist
-    if (!file.metaData?.name) {
-        if (!file.metaData) {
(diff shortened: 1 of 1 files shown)


alisa911:
The current behavior aligns with our implementation in the OsmAnd app. The 'name' tag in the metadata remains unchanged and is only added if it was previously missing.

FozzieHi:
Thanks @alisa911, so is the issue or the PR incorrect? I don't see why you would want to have the metadata name in the GPX file to be the old name. It seems to me that it isn't being updated correctly, have I just updated it in the wrong place?

dmpr0:
Hi, @FozzieHi! 
OsmAnd always show file name for clarity, so user can find imported file by displayed name. Additionally, OsmAnd try to follow the rule not to change user GPX files without necessary.

PR #393, opened 2024-05-10
Track name is not updated in GPX export xml

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/387

RZR-UA (approved):

Issue #367, opened 2024-03-27
Use system language by default


dmpr0:
## Description
Currently English is set as default language for [osmand.net/map](osmand.net/map).
* Change language priority in Chrome `chrome://settings/languages`

## ToDo
**User not logged in:**
- [ ] Use System / Browser default language for UI

**User logged in:**
- [ ] Use System / Browser language by default until user changed it

Settings
- [ ] Add new item `System (Device language)` for **Display language** option.

dmpr0:
# Review 

<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th>Chrome</th>
        <th>Safari</th>
      </tr>
      <tr>
        <td><img src="https://github.com/osmandapp/web/assets/9365308/dfd1d818-d94e-4594-82f1-01339ba7caec"  width="300" height="auto"/> </td>
        <td><img src="https://github.com/osmandapp/web/assets/9365308/e8bf0fbc-0038-49ab-8fe3-a4113e28412f"  width="300" height="auto"/> </td>
      </tr>
    </table>   
</details>   


- [ ] Use proper language names

 - - -

PR #394, opened 2024-05-10
Use system language by default

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/367

RZR-UA (approved):

PR #399, opened 2024-05-16
Add profile color for routes with undefined style

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2342

PR #427, opened 2024-06-20
Clarify that marker widget information behaviour changes depending whether user is touching the screen or not

mnalis:
As reported in https://github.com/osmandapp/OsmAnd/issues/20112, the marker widget displays different information depending if map is static, or if user is touching the screen (map panning, zooming).

This PR attempts to at least clarify the situation in the documentation (although some in-app improvements would be much better, as noted in that issue)

(also fixes some English)


EugeneZmeuk:
- [ ] [Points on the map](https://osmand.net/docs/user/personal/markers)
- [ ] [Widgets > Markers](https://osmand.net/docs/user/widgets/markers)

PR #457, opened 2024-08-16
16223

shurasmiling:


EugeneZmeuk:
for future (Favorites article), we need to check and update all screenshots

PR #464, opened 2024-08-20
Download OsmAndCollection as .travel.obf

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18866

PR #494, opened 2024-09-09
fix open route

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/443

RZR-UA (approved):

Issue #450, opened 2024-08-02
GPX track from myrouteapp has straight lines view


EugeneZmeuk:
## Description

The track was created on myrouteapp.com. [06-16 Sperlinga (3).gpx.zip](https://github.com/user-attachments/files/16469327/06-16.Sperlinga.3.gpx.zip)
I can open it on the OsmAnd app:

<img width="320" alt="image" src="https://github.com/user-attachments/assets/41a16b35-578e-48c2-9704-8f2e5477da0e">

But for OsmAnd Web:

![image](https://github.com/user-attachments/assets/2c6790ab-5700-47f0-8c7f-b7de002ed772)

I understand that this track has Route points, but when I change the Profile for all track I receive the next message

![image](https://github.com/user-attachments/assets/a2bb97ab-841b-4846-9419-25b65531e387)

But this track contains info about Track points (and the OsmAnd app show normal track and info about Route points on the Points menu)

## Steps of reproduce

1. Import the track to the OsmAnd Web
2. Try to change Profile (motorcycle) and add one more point

## Actual result

1. Straight line
2. Not create new routing for all track (there is info about Route point)

## Expected result

The same result after importing as on the OsmAnd app

vshcherb:
We should display as in Android (check the code please - probably we display trkpt instead of rtept if there are trkseg present)

PR #496, opened 2024-09-11
fix download track

alisa911:
https://github.com/osmandapp/web/issues/450


alisa911:
https://github.com/osmandapp/web/issues/450

RZR-UA (approved):

PR #525, opened 2024-10-02
Fix crash on failed download-backup request

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2714

PR #544, opened 2024-10-17
Configure map Terrain overlay

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2677

RZR-UA (approved):

Issue #545, opened 2024-10-17
Improve Wikimedia images licence parsing


dmpr0:
## Description

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/2d8f9f38-5e5e-4a1f-beb2-6f8856a85775"  width="300" height="auto"/> </td>
  </tr>
</table>   

We still can find licence in form like `Q18199165`.

* https://query.wikidata.org/

```
SELECT ?item ?itemLabel ?value WHERE {
  ?item wdt:P2479 ?value .
  SERVICE wikibase:label { bd:serviceParam wikibase:language "[AUTO_LANGUAGE],en". }
}
LIMIT 100
```

dmpr0:
# Review

<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
    <th></th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/900e889f-a31d-4550-941a-ee368a8b5346"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/ca9b8493-fc05-46cc-85c1-8a3c3f759419"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/11d93df7-28bc-4b29-a4c7-ef46a1b271c8"  width="300" height="auto"/> </td>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/b812a868-b25d-4abe-90b8-52df629d91a3"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/c60def65-0856-4fd9-9531-75feac984005"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/92a81103-e0ac-40da-9ced-8ae5a9072dc3"  width="300" height="auto"/> </td>
  </tr>
</table>  


- [ ] Different or missed data on Android
* Point [Lukyanivska Prison](https://www.openstreetmap.org/relation/6632354) – https://osmand.net/map/search/#14/50.4482/30.5026

dmpr0:
# Review 2

- [ ] 

- - -

@alisa911 Please provide details about changes, and what exactly Android (and iOS in the future) should change in the current implementation.

alisa911:
To use the updated API for retrieving image metadata, you need to:
[...]

PR #556, opened 2024-10-23
Parse licеnses

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/545

RZR-UA (changes requested):
Please move JSON and Q* processing to OsmAnd Java Server

vshcherb:
Licenses should be parsed on backend, so Android/ iOS could do it.

RZR-UA (approved):

PR #576, opened 2024-11-05
[documentation] Use @docusaurus/plugin-client-redirects to get back 404 links

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2737

Issue #539, opened 2024-10-14
Regression: Clicking on favorites on the map doesn't open context menu


dmpr0:
### How to reproduce

* Open https://osmand.net/map
* Go to **Menu** – **Configure Map**
* Turn on **Favorites** 
* Close **Configure Map**
* Try click on any favorite point

### Actual result

<table style="width:100%">
  <tr>
    <td><video src="https://github.com/user-attachments/assets/7f268717-d31c-4b04-b17b-5ee641bd817c
"  width="600" height="auto"/> </td>
  </tr>
</table>   

* Click on point doesn't do anything

### Expected result

* Clic on favorites points should open context menu

**Environment**
OsmAnd Version: OsmAnd Web, 14.10.2024 
Google Chrome

dmpr0:
# Review

[ClickLag480P.mov.zip](https://github.com/user-attachments/files/17660462/ClickLag480P.mov.zip)

- [ ] The more your click, the more you wait
- [ ] Cache?

PR #582, opened 2024-11-11
fix multiple clicks on fav markers

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/539

map/src/map/layers/FavoriteLayer.js
```diff
@@ -163,7 +163,10 @@ const FavoriteLayer = () => {
                         return;
                     }
                     layer.options.isFavorite = true;
-                    layer.on('click', onClick);
+                    if (!layer.options.hasClickHandler) {
+                        layer.on('click', onClick);
+                        layer.options.hasClickHandler = true;
+                    }
```
RZR-UA, line 169: Extra `layer.options.hasClickHandler` seems redundand.
It should be better to try useCallback instead of "function onClick()".

PR #592, opened 2024-11-18
improve waitByRemoved

alisa911:


tests/selenium/src/lib.mjs
```diff
@@ -119,11 +119,21 @@ export async function waitByRemoved(by) {
...
                 try {
                     const found = await driver.findElements(by);
-                    return !found || found.length === 0;
+                    if (!found || found.length === 0) {
+                        return true;
+                    }
+                    const element = found[0];
+                    return (
+                        (await element.getCssValue('display')) === 'none' ||
+                        (await element.getCssValue('visibility')) === 'hidden'
```
RZR-UA, line 128: Please add new method waitByRemovedOrHidden instead of modify waitByRemoved

RZR-UA (approved):

Issue #606, opened 2024-12-05
Not showing ★ button for POI chosen on the map


EugeneZmeuk:
## Description

★ button (Add to Favorites) shows only for POIs chosen from the Search menu, not from the map.

Not POI's icon for POI chosen from the searching list.

## Steps to reproduce

1. write "Castle" on the search line.
2. chosen POIi on the Map (any icon).

## Actual result 

**Chosen POI on the map**

Search line:
<img width="636" alt="image" src="https://github.com/user-attachments/assets/73d32000-caad-4851-a545-a1bacb7bd474">

Categories Search: (only here I see the right result)
<img width="663" alt="image" src="https://github.com/user-attachments/assets/8443bc73-f05b-4044-8ad3-4bf82b225b92">

**Chosen POI on the lists**

Search line:
<img width="626" alt="image" src="https://github.com/user-attachments/assets/d3cdce74-02ed-43a5-a233-862e1718c858">

Categories search:
<img width="607" alt="image" src="https://github.com/user-attachments/assets/9d6e1ec5-e63f-4c04-9384-fc4ed2f1b972">

## Expected result

1. Yes ★ button
2. there is POI's icon near POI's name

For all chosen POIs (on the map and in searching lists)

Zirochkabila:
# Testing 1

The UI now allows adding “★ ” 
However, it is impossible to check the functionality of the button due to a malfunction of the “Save” button

[Save button.zip](https://github.com/user-attachments/files/18232706/Save.button.zip)



https://github.com/user-attachments/assets/ee4e41fd-ca7c-4893-a331-764c1df43568

alisa911:
- [ ] Add error handling for the "The specified key does not exist: *** in bucket prod-user-data" error to ensure the list-files functionality does not crash on the web.

Zirochkabila:
# Testing 2
Fixed

Zirochkabila:
# Testing 2
Fixed

PR #615, opened 2024-12-12
Fix showing ★ button for all POI

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/606

PR #638, opened 2024-12-30
Update OsmAnd GPX docs: appearance, activity, waypoints, html, link, OBF collection

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2277

PR #663, opened 2025-01-31
Clarify OsmAnd GPX docs, add brief "GPX tags in UI" section

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2846

PR #669, opened 2025-02-13
Share GPX  2

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2527

map/src/menu/tracks/CloudTrackItem.jsx
```diff
@@ -69,50 +66,20 @@ export default function CloudTrackItem({ id = null, file, visible = null, isLast
...
-            }).then((newGpxFiles) => {
-                updateTracks(ctx, smartf, newGpxFiles);
-            });
-        }
-    }
-
     // Display track on map (visible or not), add to visible cache
     useEffect(() => {
         if (displayTrack === true || displayTrack === false) {
+            checkedSwitch = !checkedSwitch;
```
RZR-UA, line 72: It seems to be an incorrect way to change `checkedSwitch` inside Effect without declaring with `useState` `checkedSwitch/setCheckedSwitch.`

map/src/infoblock/components/InformationBlock.jsx
```diff
@@ -113,6 +144,15 @@ export default function InformationBlock({ showInfoBlock, setShowInfoBlock, setC
         }
     }, [showInfoBlock]);
 
+    // update URL for info track menu
+    useEffect(() => {
+        if (trackName && !ctx.shareFile) {
+            navigate(MAIN_URL_WITH_SLASH + TRACKS_URL + INFO_MENU_URL + encodeURIComponent(btoa(trackName)), {
```
RZR-UA, line 150: The `btoa/atob` may cause issues with proper `utf-8` encoding. Please look into MDN and this link: https://stackoverflow.com/questions/30106476/using-javascripts-atob-to-decode-base64-doesnt-properly-decode-utf-8-strings

RZR-UA: Possibly, it's easier to use some kind of md5sum (or other std hash) instead of btoa/atob. You even could try to use first 4-5 chars from the hash value.

RZR-UA (approved):

PR #679, opened 2025-02-24
Share GPX 2

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2527

map/src/menu/share/ShareFile.jsx
```diff
@@ -98,7 +102,10 @@ export default function ShareFile() {
                 }
             }
         }
-        fetchFile().then();
+        fetchFile().then(() => {
+            setProcessingAccess(false);
+            refreshGlobalFiles({ ctx }).then();
+        });
```
RZR-UA, line 108: It's better not to mix `then()` and `await` in `async` functions. Please prefer `await` when possible.

RZR-UA (approved):

Issue #661, opened 2025-01-30
Freezings when editing a track on the web


DmitryAlexei:
### Description
The issue is reported by several users of OsmAnd web. Track cannot be saved after editing, also crashes have been reported.

Users complains, that OsmAnd web freezes when editing a long track. The possible cause is probably in the routing (can be reproduced if the distance between added points is many km). Local storage in that case overloads quickly. 

It is proposed to resolve this issue in a complex: redo and optimize how track is stored on the site, then check the freezings. Analytics runs in parallel and should not slow down anything.

### Steps to reproduce
- create a long cycling track (more than 200 km)
- open it using OsmAnd web and start to edit it, adding points several km away from the initial track

### Actual result
- track cannot be saved due to local storage overloading
- some users experience crashes

![Image](https://github.com/user-attachments/assets/c9145e13-15f7-4921-b94c-5d57a2f5c0fc)

### ToDo
- [x] optimize local storage for long tracks
- [x] fix crashes when additional info will be provided

Zirochkabila:
# Testing 1

If the track is longer than 200 km, freeze frames start when scrolling through the elevation line


<details><summary>Freeze</summary>
<video src="https://github.com/user-attachments/assets/d5e3922d-f1b4-443d-8c0f-05d2f3d885f3"></video>
</details>

Zirochkabila:
# **Testing 2**

## 🟡 **Minor**

- [ ] **1. Smooth braking is gone - it works very fast**
- **Issue:** On long tracks (EuroVelo 6 - developed 2), when moving the line along the graph, flickering occurs, accompanied by the disappearance and appearance of a frame around the graph.

<details><summary>Details</summary>
<video src="https://github.com/user-attachments/assets/7f5aca1b-2518-4aef-9942-5fc1238fdf5c"></video>
</details>

PR #690, opened 2025-03-04
speed up slider with graph

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/661

RZR-UA (approved):

Issue #684, opened 2025-02-27
Cannot add the next point during creating a track


Zirochkabila:
### Description
When creating a track, cannot add the next point
The workaround is to press the back button (“Refresh page” or something like that) and create the track again

### Steps to reproduce

1. Go to [OsmAnd Map](https://osmand.net/map/plan/#10/47.9720/13.5187) > Plan a route > Create track > Add first and second point

### Actual result

<details><summary></summary>
<video src="https://github.com/user-attachments/assets/c6f22078-5db1-40ab-a16a-dcbd9655bd88
"></video>
</details>

> LocalClientTrackLayer.js:717 Uncaught TypeError: Cannot read properties of undefined (reading 'addLayer')
>     at t.R (LocalClientTrackLayer.js:717:36)

### Expected result

The next point is added to the map and a track is built

Zirochkabila:
# Testing 1
The ability to add the next point when creating a track has been checked
**Fixed**

PR #693, opened 2025-03-05
fix creating local tracks

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/684

RZR-UA (approved):

Issue #628, opened 2024-12-25
Change Housename display after search


Zirochkabila:
### Description
Probably related [597](https://github.com/osmandapp/web/issues/597) 
It seems that the search within Ukraine does not work

Nothing is displayed in the search for `саксаганського 69`

[OsmAnd](https://osmand.net/map/search/?pin=50.437668,30.516562#19/50.43671/30.50533)
[OsmAnd test](https://test.osmand.net/map/search/#19/50.43672/30.50515)
[OSM](https://www.openstreetmap.org/search?query=%D1%81%D0%B0%D0%BA%D1%81%D0%B0%D0%B3%D0%B0%D0%BD%D1%81%D1%8C%D0%BA%D0%BE%D0%B3%D0%BE%2069#map=19/50.436675/30.505404)

```
<html>
<head><title>404 Not Found</title></head>
<body>
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.18.0 (Ubuntu)</center>
</body>
</html>
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
```
```
 try {
        const fullOptions = Object.assign({}, { redirect: 'manual' }, options);
        response = await fetch(fullURL, fullOptions);
    } catch (e) {
        // got general error (have no response)
        console.debug('fetch-catch-error', url, e);
        const ret = { ok: false, text: () => null, json: () => null, blob: () => null, data: null };
        if (options?.throwErrors) {
            const error = new Error('fetch-catch-error');
            error.response = ret;
            throw error;
        } else {
            return ret;
        }
    }
```

### Steps to reproduce

1. Click on one of the links above > Type `саксаганського 69` in the search bar > Press Enter

### Actual result

<details><summary>OsmAnd & OsmAnd test</summary>
<img width="800" alt="Screenshot 2023-01-31 at 11 19 41" src="https://github.com/user-attachments/assets/45ab90c0-dc72-4ea8-929a-a717976cd8f9">
<img width="800" alt="Screenshot 2023-01-31 at 11 19 41" src="https://github.com/user-attachments/assets/475d4b62-8c05-46f9-930c-ea9a7920bd32">
[...]

PR #736, opened 2025-04-02
fix search result info

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/628

RZR-UA (approved):

Issue #711, opened 2024-12-10
Support track appearance on the map


Rowin63:
### Description

though the same settings, two tracks (see example tracks) are displayed different in the app and online.

In the app (+4.9.9):

![Screenshot_20241210_100119_OsmAnd+](https://github.com/user-attachments/assets/e88c684a-d895-4eb3-9e46-f22505aa4bde)

The same tracks online:

![2024-12-10 online track appearance](https://github.com/user-attachments/assets/74b5aaa2-0bf2-44c9-8a06-754f72a61321)

### Steps to reproduce

I dont know, in the app both tracks do have the same settings

### Actual result

two tracks with the same settings > have online two different colors 

anyway, the track´s app settings don´t "move" to the cloud (like "switched off" start & finish icons, they appear again and fill up the screen)

### Expected result

same color, same settings

### Your Environment (required)

OsmAnd Version: +4.9.9

dmpr0:
## ToDo
Tracks on the should be rendered using track appearance tags:
* Width
* Color
* Start / Finish icons

Rowin63:
So it is supposed to work correctly? 
Right now it does not. 
On the web map all my tracks still are displayed in the same colour.

Zirochkabila:
# Testing 1

The only thing that differs is the width of the line (both when importing from a cloud and when importing from a device - Import tracks button on the WEB)


<details><summary>Details</summary>
<table style="width:100%">
  <tr>
    <th>Original</th>
    <th>From Cloud</th>
    <th>From Device</th>
  </tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/c85897aa-75ad-4424-a211-84df16b64354"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/bd20f780-76b3-4de0-8a83-32b50ed3f502"  width="500" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/c4b3a009-6198-4f0d-932c-4a3ba3714098"  width="500" height="auto"/> </th>
  </tr>
</table> 
</details> 


<details><summary>Details</summary>
<table style="width:100%">
  <tr>
    <th>Original</th>
    <th>From Cloud</th>
    <th>From Device</th>
  </tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/1ff6d58e-8b57-42e7-bd29-d5e48ffcd75b"  width="300" height="auto"/> </th>
[...]

PR #738, opened 2025-04-02
fix track weight

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/711

alisa911:
https://github.com/osmandapp/web/issues/711

PR #754, opened 2025-04-14
update subs

alisa911:


map/src/login/FastSpringPurchaseButton.jsx
```diff
@@ -15,10 +16,30 @@ export default function FastSpringPurchaseButton({ selectedProducts, testMode =
...
+                        if ((info?.valid === 'true' && info?.startTime && info?.expireTime) || attempt >= 5) {
+                            testMode && console.log('✅ Updated info after payment');
+                        } else {
+                            setTimeout(() => tryUpdate(attempt + 1), 3000);
+                        }
+                    });
+                };
+                tryUpdate();
+            }
+        };
```
RZR-UA, line 42: 1. `5 attempts * 3 seconds` is not enough (think about 3DS payment confirmation, which could take even 10+ minutes). Please consult with the FS guys about the actual wait-for-result timeout.
2. Try to emulate the Websockets approach using a "long-waiting" API request. Server-side pseudo-code: `while(result || timeout) { result=query(); sleep(1); } ... return result;`; timeout should be at least 60 seconds (but anyway less than 300 seconds, which is the default http-request timeout).

RZR-UA (approved):

PR #778, opened 2025-04-29
remove zoom from get-wiki-data api

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2964

RZR-UA (approved):

PR #802, opened 2025-05-09
Add fr locale to docusaurus (draft for tests)

RZR-UA:


RZR-UA:
To Fix

- [ ] Click on Map menu (/map) broken: `https://test.osmand.net/fr/map` = 404 from [fr page](https://test.osmand.net/fr/docs/user/purchases/)

Gateway07:
> To Fix
> * [ ]  Click on Map menu (/map) is broken: `https://test.osmand.net/fr/map` = 404 from [fr page](https://test.osmand.net/fr/docs/user/purchases/)
There is known limitation of Docusaurus development server (https://dongjunyang.netlify.app/docs/web/3/):
You can not test all the language at once. It means that you can not test all the language with yarn run start or npm run start. So you need to test with below command.

PR #823, opened 2025-06-05
Restore lost files from shurasmiling commits

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2903

PR #863, opened 2025-06-20
Fix map jitter / URL hash loop

oleks4ndr:
Issue:
- When manually changing URL hash coordinates by ± 0.1 or 0.01, the map would occasionally enter a loop of jitter / shaking (url bouncing between two near identical cords). 

Fix:
- Remove the extra Leaflet Hash instance.

### DIFF
--- a/map/src/map/OsmAndMap.jsx
+++ b/map/src/map/OsmAndMap.jsx
@@ -55,19 +55,17 @@ const OsmAndMap = ({ mainMenuWidth, menuInfoWidth }) => {
     const attributionSize = 300;
     const marginLeft = width / 2 - attributionSize + menuMargin;
 
-    const hashRef = useRef(null);
-
     const whenReadyHandler = (event) => {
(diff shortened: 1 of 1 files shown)


map/src/map/OsmAndMap.jsx
```diff
@@ -55,17 +55,19 @@ const OsmAndMap = ({ mainMenuWidth, menuInfoWidth }) => {
     const attributionSize = 300;
     const marginLeft = width / 2 - attributionSize + menuMargin;
 
+    const hashRef = useRef(null);
+
     const whenReadyHandler = (event) => {
         const { target: map } = event;
-        if (map) {
-            new L.Hash(map);
+        if (map && !hashRef.current) {
```
RZR-UA, line 62: If whenReadyHandler is called once, useRef is not required. But if it is called multiple times, your code breaks the inner block of this if (mapRef.current is not updated, etc). Could you try to refactor in a better way?

PR #872, opened 2025-06-23
Detect available languages of web-translation.json (PricingPage)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/web/pull/873

Issue #765, opened 2025-04-18
Show city names for search results


dmpr0:
## ToDo

Show city name on the second line for search result.

<table style="width:100%">
  <tr>
    <th>Web</th>
    <th>Android</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/bb3232a3-231e-46e1-833e-9a327202f953"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/c2e2aa08-cc38-4e2f-b5b2-c436db1c1a8b"  width="300" height="auto"/> </td>
  </tr>
</table>   

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/71fe1ca3-84b5-4951-8f74-5cf3f0687c72"  width="300" height="auto"/> </td>
  </tr>
</table>

alisa911:
<img width="430" alt="Image" src="https://github.com/user-attachments/assets/93badc24-1ad4-40ab-a8ad-990a8531d92a" />

Zirochkabila:
# Testing 1

Checked the display of the city in the second line of the search result - **Added**

dmpr0:
## UI Review

- [ ] Direction arrow should be removed or logic should be added.
- - [ ] No location > grey / from map center
- - [ ] Location > blue > from location

PR #879, opened 2025-06-30
Add logic to directional arrow in search results (https://github.com/osmandapp/web/issues/765)

oleks4ndr:
- Add util to calculate angle between two points of coordinates
- Add logic to search results component for directional arrow: 
    - debounced lat/lon changes (300ms) to avoid excessive recalculations when moving the map
    - pass bearing and type of user location (geolocation or center of map/hash) down to child
- SearchResultItem component:
    - render arrow and distance blue is geolocation is enabled in browser by user
    - not enabled location service => arrow and distance grey 
    - if bearing is undefined default to no rotation.
    
- issue: https://github.com/osmandapp/web/issues/765#issuecomment-2893821985

### DIFF
--- a/map/src/menu/search/search/SearchResults.jsx
+++ b/map/src/menu/search/search/SearchResults.jsx
@@ -10,7 +10,7 @@ import Loading from '../../errors/Loading';
 import { useGeoLocation } from '../../../util/hooks/useGeoLocation';
 import { LOCATION_UNAVAILABLE } from '../../../manager/FavoritesManager';
 import { getCenterMapLoc } from '../../../manager/MapManager';
-import { getDistance } from '../../../util/Utils';
+import { getDistance, getBearing } from '../../../util/Utils';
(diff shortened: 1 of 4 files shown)


map/src/menu/search/search/SearchResults.jsx
```diff
@@ -50,8 +50,25 @@ export default function SearchResults({ value, setOpenSearchResults, setIsMainSe
...
+
+    useEffect(() => {
+        const handler = setTimeout(() => {
+            setDebouncedLatLon({ lat, lon });
+        }, 300);
+
+        return () => {
+            clearTimeout(handler);
+        };
+    }, [lat, lon]);
```
RZR-UA, line 64: Weird debouncer code. Please PM me.

RZR-UA: Seems ok.

RZR-UA (approved):

PR #912, opened 2025-07-22
fix removing old track layers from map

alisa911:


map/src/map/util/MapManager.js
```diff
@@ -11,3 +11,10 @@ export function getVisibleBbox(map, ctx) {
 
     return L.latLngBounds(topLeft, bottomRight);
 }
+
+export function addLayerToMap(map, layer, id) {
+    if (layer && !map.hasLayer(layer)) {
+        layer.options = { ...layer.options, id };
+        map.addLayer(layer);
+    }
```
RZR-UA, line 19: Please add `else` with `console.debug('avoid adding map layer', id)`

RZR-UA (approved):

Issue #921, opened 2025-07-23
Explore - When changing the zoom level, markers stop appearing on the map


alisa911:
When changing the zoom level from small, where the explore is disabled, to large, the explore objects no longer display on the map and context menu.
The progress bar endlessly in the context menu.

<img width="1318" height="741" alt="Image" src="https://github.com/user-attachments/assets/aff745b3-1a86-441c-836b-78311dca2496" />

Zirochkabila:
# Testing 1

Checked the display of Explore points at different zooms (When changing the zoom level) - **Fixed**

PR #922, opened 2025-07-23
fix empty explore objs

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/921

RZR-UA (approved):

PR #924, opened 2025-07-24
Add upload kml, kmz file

Dima-1:


RZR-UA (approved):

map/src/frame/util/LocalGpxUploader.jsx
```diff
@@ -50,7 +50,7 @@ export default function LocalGpxUploader({ children }) {
 
     return (
         <label htmlFor="se-upload-gpx">
-            <HiddenInput id="se-upload-gpx" accept=".gpx" multiple type="file" onChange={fileSelected} />
+            <HiddenInput id="se-upload-gpx" accept=".gpx, .kmz, .kml" multiple type="file" onChange={fileSelected} />
```
RZR-UA, line 53: If Kml2Gpx is unable to convert KML geometry, please accept KML/KMZ files only in Import Favorites. Add boolean parameter acceptKmlKmz to the component.

RZR-UA (changes requested):

RZR-UA (approved):

Issue #857, opened 2025-06-19
Missing names: for Popular Places (Wikidata) with no POI on the map


DmitryAlexei:
### Description
Some Popular Places on the map don't have Wikidata links and Names in context menu. 

### ToDo
- [x] Add Wikidata links when missing
- [x] Investigate why names are missing

## Example 1
* Site link [42.051920, -70.185250](https://osmand.net/map/search/#18/42.05187/-70.18511) 
* [Wikidata](https://www.wikidata.org/wiki/Q116632800)

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/0c0fa9c8-80cd-4f61-8998-0673aa6250c8"  width="400" height="auto"/> </td>
  </tr>
</table>

## Example 2
* Site link [42.057430, -70.178930](https://osmand.net/map/search/#19/42.05720/-70.17905)

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/bf36e13e-7fa4-47ae-8cf9-4703a79847fb"  width="600" height="auto"/> </td>
  </tr>
</table>

## Example 3

<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/6edecf5d-4fc1-4a6e-9683-d6635467ec8d"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/58e32ed9-c6d3-4d10-9742-c135321dbf05"  width="300" height="auto"/> </td>
  </tr>
</table>   

* Coordinates – `50.432884, 30.527348`
* Link – https://osmand.net/map/search/#17/50.43289/30.52832

## Example 4

<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/eb548056-3265-4b82-bd98-2c16c007d48a"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/f8e010fe-70cf-41c0-8533-9fd776b49adb"  width="300" height="auto"/> </td>
  </tr>
</table>   

* OsmAnd https://osmand.net/map/search/#17/50.43289/30.52832
* OSM – http://openstreetmap.org/way/283675449#map=17/50.432108/30.532100
* Wikidata – https://www.wikidata.org/wiki/Q122296458

alisa911:
[...]

PR #927, opened 2025-07-25
Add Wikidata to POI when missing

oleks4ndr:
- Add Wikidata to tags when currentPoi is undefined

https://github.com/osmandapp/web/issues/857

### DIFF
--- a/map/src/infoblock/components/wpt/WptDetails.jsx
+++ b/map/src/infoblock/components/wpt/WptDetails.jsx
@@ -246,8 +246,6 @@ export default function WptDetails({ isDetails = false, setOpenWptTab, setShowIn
                     });
                     fallbackTags.push(wikidataTag);
                     tags = { res: fallbackTags };
-                } else {
-                    tags = null;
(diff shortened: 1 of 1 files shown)


map/src/infoblock/components/wpt/WptDetails.jsx
```diff
@@ -231,7 +235,20 @@ export default function WptDetails({ isDetails = false, setOpenWptTab, setShowIn
                 }
             } else if (type?.isSearch || type?.isPoi || type?.isWikiPoi) {
                 const currentPoi = ctx.selectedWpt.poi;
-                tags = currentPoi ? await WptTagsProvider.getWptTags(currentPoi, type, ctx) : null;
+                if (currentPoi) {
+                    tags = await WptTagsProvider.getWptTags(currentPoi, type, ctx);
+                } else {
+                    const fallbackTags = [];
```
alisa911, line 241: No need to create this empty array if the object doesn't contain wikidata

PR #932, opened 2025-08-04
docusaurus-opt-fix

Gateway07:
1. Install the @docusaurus/faster dependency 
2. Change docusaurus.config.js

https://github.com/osmandapp/OsmAnd-Issues/issues/2903


RZR-UA:
40GB is unacceptable for the current servers

Gateway07 (commented):
Looks good for me.

PR #936, opened 2025-08-07
keep tracks and favorites menu states after switching menus

alisa911:


alisa911:
Screenshot 2025-08-26 at 13 55 07
Screenshot 2025-08-26 at 13 55 16

map/src/App.js
```diff
@@ -54,51 +54,79 @@ const NavigateGlobal = () => {
...
-                        <Route path={MAIN_URL} element={<GlobalFrame />}>
-                            <Route path={LOGIN_URL} element={<LoginMenu />}>
-                                <Route path={PURCHASES_URL} element={<PurchasesMenu />}>
-                                    <Route path=":key" element={<PurchaseInfo />}></Route>
+        <ResetAppContext.Provider value={resetApp}>
+            <LoginContextProvider key={`login-${resetKey}`}>
+                <AppContextProvider key={`app-${resetKey}`}>
+                    <BrowserRouter
+                        key={`router-${resetKey}`}
+                        future={{ v7_startTransition: true, v7_relativeSplatPath: true }}
```
RZR-UA, line 86: Please explain the magic in comments.

map/src/manager/track/TracksManager.js
```diff
@@ -1373,6 +1373,7 @@ export async function openTrackOnMap({
                 newGpxFiles[ctx.selectedGpxFile.name].url = null;
             }
         }
+        console.log(showOnMap);
```
RZR-UA, line 1376: drop log

map/src/map/OsmAndMap.jsx
```diff
@@ -58,8 +58,11 @@ const OsmAndMap = ({ mainMenuWidth, menuInfoWidth }) => {
     const whenReadyHandler = (event) => {
         const { target: map } = event;
         if (map) {
-            window.seleniumTestsMap = map;
             const hash = new L.Hash(map);
+
+            window.__leafletMap = map;
```
RZR-UA, line 63: Could you remove global leaflet refs if possible?

map/src/map/layers/FavoriteLayer.js
```diff
@@ -317,9 +316,9 @@ const FavoriteLayer = () => {
                 ? e.sourceTarget.options.category
                 : FavoritesManager.DEFAULT_GROUP_NAME;
             ctx.selectedGpxFile.id = e.sourceTarget.options.groupId;
+            ctx.selectedGpxFile.key = `${e.sourceTarget.options.groupId}:${ctx.selectedGpxFile.markerCurrent.name}`;
```
RZR-UA, line 319: Duplication in usage of ctx.selectedGpxFile.markerCurrent.name and e.sourceTarget.options.groupId. Please use just set fields: `ctx.selectedGpxFile.id + ':' + ctx.selectedGpxFile.name`

map/src/menu/actions/GroupHeader.jsx
```diff
@@ -60,8 +60,22 @@ export default function GroupHeader({
...
+        if (type === TRACKS_TYPE) {
+            ctx.openGroups.pop();
+            ctx.setOpenGroups([...ctx.openGroups]);
+        } else if (type === FAVORITES_TYPE) {
+            ctx.openFavGroups.pop();
+            ctx.setOpenFavGroups([...ctx.openFavGroups]);
+        }
+    }
+
+    function getGroupLength() {
```
RZR-UA, line 72: getGroupsCount?

map/src/manager/GlobalManager.js
```diff
@@ -18,7 +18,7 @@ export const MAIN_PAGE_TYPE = 'main';
 export const MAIN_URL = '/map';
 export const MAIN_URL_WITH_SLASH = '/map/';
 export const SEARCH_URL = 'search/';
-export const EXPLORE_URL = 'search/explore/';
+export const EXPLORE_URL = 'explore/';
```
RZR-UA, line 21: Please comment that EXPLORE_URL always depends on SEARCH_URL.

tests/selenium/src/lib.mjs
```diff
@@ -162,25 +162,52 @@ export async function clickBy(by, { optional = false } = {}) {
...
+                        const r = (await c.getRect?.().catch(() => ({}))) || {};
+                        if (vis && (r.width ?? 0) > 0 && (r.height ?? 0) > 0) {
+                            target = c;
+                            break;
+                        }
+                    }
+                }
+            } catch (_) {
+                // ignore visibility probing errors; fallback to original element
+            }
```
RZR-UA, line 183: This is a library code. You should never modify library behavior in such a way. Compatibility is sacred.
1. If you really can not do without this code, move it into a method and call the method if a new flag is specified, e.g., findFirstVisible === true.
2. All in all, it is better to avoid Elements with non-unique IDs in the DOM tree.

RZR-UA (changes requested):

tests/selenium/src/tests/26-map-select-marker.mjs
```diff
@@ -129,6 +145,24 @@ async function testMarkers(coords1, coords2, coords3, indexes = [0, 1], type) {
     }
 }
 
+export async function waitUntilNotSkeleton(by) {
```
RZR-UA, line 148: remove this unknown technology

map/src/util/hooks/useIsVisible.js
```diff
@@ -0,0 +1,57 @@
+import { useLayoutEffect, useRef, useState } from 'react';
```
RZR-UA, line 1: This approach might be slow. It is overkill to use a cycle `while(parent)` for each affected button in each re-render. Try to simplify, for instance, use `trackGroup+favoriteGroup` as an additional part of                             `id={`se-back-folder-button-${isVisible}`}`, etc.

PR #950, opened 2025-08-12
Fix cloud kml 2

Dima-1:


map/src/frame/util/CloudGpxUploader.jsx
```diff
@@ -55,7 +62,11 @@ export default function CloudGpxUploader({ children, folder = null, style = null
                     ctx.setTrackLoading([...ctx.trackLoading.filter((n) => n !== file.name)]);
                 }
             });
-            reader.readAsText(file);
+            if (file.name.toLowerCase().endsWith('.kmz')) {
```
RZR-UA, line 65: No hardcode for kml/kmz/gpx. Please use KML_FILE_EXT, KMZ_FILE_EXT.

map/src/manager/track/TracksManager.js
```diff
@@ -1274,7 +1275,9 @@ export function getGpxFiles(listFiles) {
     return (!listFiles || !listFiles.uniqueFiles ? [] : listFiles.uniqueFiles).filter((item) => {
         return (
             (item.type === 'gpx' || item.type === 'GPX') &&
-            (item.name.slice(-4) === '.gpx' || item.name.slice(-4) === '.GPX' || item.name.endsWith(EMPTY_FILE_NAME))
+            (item.name.slice(-4) === GPX_FILE_EXT ||
```
RZR-UA, line 1278: lowercase instead of .GPX ?

PR #1061, opened 2025-10-02
fix poi config layer

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1063

RZR-UA (approved):

PR #1118, opened 2025-10-16
Add section reverse track

TinaPohor:
https://github.com/osmandapp/OsmAnd/issues/20812


EugeneZmeuk (commented):
Confirmation from Victor

PR #1133, opened 2025-10-20
fix osm url, use get poi api only for pois

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd/issues/22642

PR #1218, opened 2025-11-13
Add navigation click

alisa911:


map/src/map/OsmAndMap.jsx
```diff
@@ -48,9 +49,24 @@ export function addClicksToMap(map, ctx) {
...
+        if (!globalThis.location.pathname.includes(NAVIGATE_URL)) {
+            return false;
+        }
+        const routeObject = ctx.routeObject;
+        if (!routeObject) {
+            return false;
+        }
+        const startPoint = routeObject.getOption(ROUTE_POINTS_START);
+        const finishPoint = routeObject.getOption(ROUTE_POINTS_FINISH);
+        return !startPoint || !finishPoint;
```
RZR-UA: Please move this code to a place closer to Navigation (NavigationManager or GlobalManager)

Issue #1210, opened 2025-11-08
Reordering the points of a track crashes the map


kabouter:
When I create a track and try to reorder a point in the list by dragging it to the correct position, OsmAnd crashes.

When I reload the page and open the route/track again or try to display it on the map with the toggle, OsmAnd crashes again.

https://github.com/user-attachments/assets/cc01a739-65cc-4489-89d3-b8f6b2a6fb6d

The console errors:

```

TypeError: i.geometry.forEach is not a function
    l TrackLayerProvider.js:158
    Kx TrackLayerProvider.js:64
    Kx TrackLayerProvider.js:62
    zY LocalTrackLayerHelper.js:58
    Y LocalClientTrackLayer.js:490
    Y LocalClientTrackLayer.js:192
    React 7
    w scheduler.production.min.js:13
    T scheduler.production.min.js:14
    7234 scheduler.production.min.js:14
    Webpack 10
react-dom.production.min.js:188:120
    React 13
    w scheduler.production.min.js:13
    T scheduler.production.min.js:14
    (Async: EventHandlerNonNull)
    7234 scheduler.production.min.js:14
    Webpack 10


TypeError: e.geometry.forEach is not a function
    hW TracksManager.js:298
    hW TracksManager.js:296
    C9 GpxGraphProvider.jsx:150
    C9 GpxGraphProvider.jsx:164
    React 7
    w scheduler.production.min.js:13
    T scheduler.production.min.js:14
    7234 scheduler.production.min.js:14
    Webpack 10
react-dom.production.min.js:188:120

Uncaught TypeError: i.geometry.forEach is not a function
    l TrackLayerProvider.js:158
    Kx TrackLayerProvider.js:64
    Kx TrackLayerProvider.js:62
    zY LocalTrackLayerHelper.js:58
    Y LocalClientTrackLayer.js:490
    Y LocalClientTrackLayer.js:192
    React 7
    w scheduler.production.min.js:13
    T scheduler.production.min.js:14
    7234 scheduler.production.min.js:14
    Webpack 10
TrackLayerProvider.js:158:20
```

PR #1254, opened 2025-11-25
fix crash with points reorder

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1210

RZR-UA (approved):

PR #1302, opened 2025-12-11
Navigation UI

alisa911:


map/src/util/hooks/navigation/useNavigationHistory.js
```diff
@@ -1,86 +1,106 @@
...
+import { useCallback, useEffect } from 'react';
 import { ROUTE_POINTS_START, ROUTE_POINTS_FINISH, ROUTE_POINTS_VIA } from '../../../store/geoRouter/profileConstants';
 import { FINISH_POINT, INTERMEDIATE_POINT, START_POINT } from '../../../menu/navigation/NavigationInputRow';
+import { navigationObject } from '../../../store/navigationObject/navigationObject';
 
-export default function useNavigationHistory(routeObject) {
-    const [history, setHistory] = useState([]);
+export default function useNavigationHistory(navObject, ctx = null) {
+    const history = ctx.navigationHistory || [];
+    const setHistory = ctx.setNavigationHistory;
```
RZR-UA: NPE ctx = null

map/src/util/hooks/navigation/useNavigationHistory.js
```diff
@@ -1,86 +1,106 @@
...
+                    poiType: navObj.poiType,
+                    icon: navObj.icon,
+                    displayValue: navObj.displayValue,
+                    usedAt: navObj.usedAt,
+                });
+
+                return [newObj, ...filtered].slice(0, 100);
+            });
+        },
+        [setHistory]
```
RZR-UA: useCallback (as well as useEffect) should not depend on the setter function

map/src/services/RouteService.js
```diff
@@ -73,7 +73,10 @@ export function RouteService() {
                 obj['end'] = finishPoint.lat.toFixed(6) + ',' + finishPoint.lng.toFixed(6);
             }
             if (viaPoints?.length > 0) {
-                obj['via'] = viaPoints.map((i) => i.lat.toFixed(6) + ',' + i.lng.toFixed(6)).join(';');
+                obj['via'] = viaPoints
+                    .filter((i) => i != null)
```
RZR-UA: Why could a null value happen in this array?

map/src/map/layers/NavigationLayer.js
```diff
@@ -104,7 +104,20 @@ const NavigationLayer = ({ geocodingData, region }) => {
             // remove focus from all inputs
             globalThis.dispatchEvent(new Event('nav-blur'));
 
-            routeObject.setOption(target, point);
+            // Handle intermediate point
+            if (target.type === ROUTE_POINTS_VIA) {
+                const viaPoints = routeObject.getOption(ROUTE_POINTS_VIA) || [];
+                const newViaPoints = [...viaPoints];
+                while (newViaPoints.length <= target.index) {
+                    newViaPoints.push(null);
```
RZR-UA: Avoid using null in the viaPoints array if possible.

RZR-UA (approved):

map/src/map/layers/NavigationLayer.js
RZR-UA: ok, keep null as a sign of empty interpoint

PR #1313, opened 2025-12-18
Fix checking result.

Dima-1:


map/src/manager/FavoritesManager.js
```diff
@@ -287,17 +287,16 @@ export async function updateFavGroupPinned({ group, updatedPointsGroups, groupNa
...
     const result = await updateFavoriteGroup(favGroupData, group, groupName);
-    if (!result) {
+    if (!result?.data?.pointsGroups?.[groupName]) {
         return;
     }
 
     ctx.setFavorites((prev) => {
         const groups = prev.groups.map((g) => {
             if (g.id === group.id) {
-                const newPinned = result.data?.pointsGroups?.[groupName]?.pinned;
```
alisa911: const newPinnedGroup = result.data?.pointsGroups?.[groupName];

Issue #1126, opened 2025-10-17
Not all small objects, such as benches, are visible on the POI layer


dmpr0:
### How to reproduce
* Open map at https://osmand.net/map/search/result/?query=Bench&type=category&key=bench#20/48.83200/12.95684

### Actual result

<table style="width:100%">
  <tr>
    <td><video src="https://github.com/user-attachments/assets/a743c576-7995-416e-a159-eebcbe11538e"  width="300" height="auto"/> </td>
  </tr>
</table>   

* Only one Bench POI is visible depends on map position

### Expected result

* All three POI's are visible
* * https://www.openstreetmap.org/way/1151422322

**Environment**
OsmAnd Version: from 17.10.2025

PR #1340, opened 2026-01-02
use new poi search for all cases

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1126

map/src/menu/search/search/SearchResults.jsx
```diff
@@ -94,15 +111,22 @@ export default function SearchResults() {
     useEffect(() => {
         if (params.query) {
             document.title = params.query;
+        } else if (params.type) {
+            const brandInfo = parseBrandType(params.type);
+            if (brandInfo) {
+                document.title = brandInfo.brandName;
+            } else {
+                document.title = getCategoryName(params.type, t, getFirstSubstring);
```
RZR-UA: Not the best idea to modify document.title. You'll never get it back. At least, use a custom hook to modify the title temporarily according to the state (search result page or other page, for example).

Issue #712, opened 2025-03-24
Waypoint Visibility Resets After Page Refresh or Logout


Dane-M-652:
Steps to reproduce:

1. Make a track containing waypoints visible
2. Toggle visibility of its waypoints to "off" using the switch on the waypoints pane of the track
3. Waypoints are hidden from map
4. Logout from Osmand Web
5. Login to Osmand Web
6. Waypoints are shown again on map (visibility toggle is unintentionally moved to "on")

Expected behavior: Waypoint visibility status is retained after login (like the visibility status of the entire track).

Waypoint visibility status is also lost if Osmand Web is just refreshed in browser (pressing F5), without logout and login.

Dane-M-652:
Waypoint visibility status is also lost if the corresponding track is selected in the track list.

alisa911:
**Dev notes (task is blocked)** 
This isn’t possible without enabling track editing mode.
Editing mode creates a local track that can be modified and is only saved to the database after confirmation.
Bypassing the existing logic to inject track changes is a bad practice. There must be a clear workflow of actions for the user.
@dmpr0 @vshcherb

vshcherb:
@alisa911 that setting is part of the track appearance, it could be implemented once Track Appearance will be implemented. It works the same as track color settings or display start/finish points

Zirochkabila:
# Testing 1

> Web - prod

### **Severity**: 🟠 Major / **Priority**: 🔴 High

- [x] 1. Tested both after refreshing the page and after logging back into the account - the waypoints are displayed as enabled again

<details><summary>Details</summary>
<video src="https://github.com/user-attachments/assets/a8c68063-aa8a-4709-b3bf-549a5a9d6b69"></video>
</details>

Dima-1:
- [x] Fix new info files: add at the beginning of the JSON {"type":"GPX","file":fileName,"subtype":"gpx"
 - [x] Create a script to fix existing info files that contain point groups only. Also add {"type":"GPX","file":fileName,"subtype":"gpx".

Dima-1:
Bug - after changing the visibility of the wpt of the track, when reopening this track without refreshing the page, the change is lost
- [x] cloud
- [x] local

Zirochkabila:
# Testing 2

> Web - prod

Tested that the display status of waypoints is preserved after refreshing the page and panning the map, as well as when reopening the track - **Fixed** - Waypoint status is preserved

PR #1522, opened 2026-03-12
Save visibility of wpt groups

Dima-1:
https://github.com/osmandapp/web/issues/712

### DIFF
--- a/map/src/infoblock/components/tabs/WaypointsTab.jsx
+++ b/map/src/infoblock/components/tabs/WaypointsTab.jsx
@@ -41,8 +41,8 @@ const WaypointGroup = ({
     }, [massOpen]);
 
     useEffect(() => {
-        mounted && setVisible(massVisible);
-    }, [massVisible]);
+        setVisible(defaultVisible);
+    }, [defaultVisible, massVisible]);
 
     const point = points[0].wpt;
     const iconHTML = createPoiIcon({ point, color: point.color, background: point.background, icon: point.icon })
@@ -269,14 +269,20 @@ export default function WaypointsTab() {
         return groups;
     }
 
-    const [showMass, setShowMass] = useState(false);
-    const [massOpen, setMassOpen] = useState(false);
-    const [massVisible, setMassVisible] = useState(() => {
+    function isMassVisible() {
         const pointsGroups = getResolvedPointsGroups(ctx.selectedGpxFile);
         const groupKeys = Object.keys(pointsGroups || {});
-        return groupKeys.length > 0 && groupKeys.every((g) => isWptGroupShown(pointsGroups, g));
-    });
+        return groupKeys.length > 0 && groupKeys.some((g) => isWptGroupShown(pointsGroups, g));
+    }
+
+    const [showMass, setShowMass] = useState(false);
+    const [massOpen, setMassOpen] = useState(false);
+    const [massVisible, setMassVisible] = useState(isMassVisible());
     const debouncerTimer = useRef(null);
+
+    useEffect(() => {
+        setMassVisible(isMassVisible());
+    }, [ctx.selectedGpxFile?.info?.pointsGroups]);
 
     const switchMassOpen = () => setMassOpen(!massOpen);
     const switchMassVisible = () => {


map/src/infoblock/components/tabs/WaypointsTab.jsx
```diff
@@ -9,16 +9,55 @@ import { confirm } from '../../../dialogs/GlobalConfirmationDialog';
 import { useWindowSize } from '../../../util/hooks/useWindowSize';
 import { createPoiIcon } from '../../../map/markers/MarkerOptions';
 import isEmpty from 'lodash-es/isEmpty';
+import { debouncer } from '../../../context/TracksRoutingCache';
+import { saveInfoFile } from '../../../manager/track/SaveTrackManager';
+
+const SAVE_DEBOUNCE_MS = 1000;
```
alisa911: Why use a constant if it is used only once?

map/src/infoblock/components/tabs/WaypointsTab.jsx
```diff
@@ -9,16 +9,55 @@ import { confirm } from '../../../dialogs/GlobalConfirmationDialog';
...
+
+function updateGroupsVisibility(ctx, groupNames, hidden, debouncerTimer) {
```
alisa911: Create a universal editor for track wpt groups. In the future, we'll also edit other parameters.

map/src/manager/track/SaveTrackManager.js
```diff
@@ -327,6 +326,40 @@ export async function saveEmptyTrack(folderName, ctx) {
     }
 }
 
+export async function saveInfoFile(selectedGpxFile) {
```
alisa911: A universal method and approach are needed for working with the info file. Its reading and editing should be moved into a separate file.

alisa911: Replacing the entire .info file on every small change (full PUT cycle) wastes traffic, causes race conditions when two clients edit different fields simultaneously, and adds unnecessary storage churn. Instead, send only the changed fields to the server and let it merge them into the existing .info — less traffic, no conflicts.

map/src/infoblock/components/tabs/WaypointsTab.jsx
```diff
@@ -317,6 +381,7 @@ export default function WaypointsTab() {
         const groups = getSortedGroups();
         const keys = Object.keys(groups);
         const trackName = ctx.selectedGpxFile.name;
+        const pointsGroups = ctx.selectedGpxFile.info?.pointsGroups ?? ctx.selectedGpxFile.pointsGroups;
```
alisa911: Remove code duplication and create a separate method with a descriptive name for ctx.selectedGpxFile.info?.pointsGroups ?? ctx.selectedGpxFile.pointsGroups с

map/src/infoblock/components/tabs/WaypointsTab.jsx
```diff
@@ -9,16 +9,72 @@ import { confirm } from '../../../dialogs/GlobalConfirmationDialog';
...
+        const updatedGpxFile = {
+            ...prevFile,
+            info: {
+                ...prevFile.info,
+                pointsGroups: updatedPointsGroups,
+            },
+        };
+
+        if (isCloudTrack(ctx)) {
+            debouncer(
```
alisa911: ctx and updatedGpxFile are captured by closure at call time, but the debouncer fires 500ms later. By then ctx.listFiles or the selected file may have changed, leading to a stale upload.

map/src/infoblock/components/tabs/WaypointsTab.jsx
```diff
@@ -302,9 +358,17 @@ export default function WaypointsTab() {
     const [showMass, setShowMass] = useState(false);
     const [massOpen, setMassOpen] = useState(false);
     const [massVisible, setMassVisible] = useState(true);
```
alisa911: massVisible always starts as true, ignoring persisted hidden state. After reload, the toggle button will be out of sync with actual group visibility.

map/src/map/util/TrackLayerProvider.js
```diff
@@ -516,6 +518,13 @@ function parseWpt({
                     };
                     ctx.setSelectedWpt(wpt);
                 });
+                marker.on('add', (e) => {
+                    const pointsGroups = data?.info?.pointsGroups || data?.pointsGroups;
```
alisa911: data is captured once when the marker is created. The add event won't re-fire for already-added markers, so toggling visibility won't affect existing markers on the map.

alisa911 (approved):

PR #1628, opened 2026-04-14
fix updating info after refresh

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/712

map/src/context/AppContext.js
```diff
@@ -110,6 +111,44 @@ function createInteractiveMap(data, type) {
     return interactiveMap;
 }
 
+function applyRefreshedInfoFilesToGpx(updatedData, setGpxFiles, setSelectedGpxFile, selectedGpxFile) {
```
RZR-UA: Consider moving this method to useInitialFilesLoad.js to keep AppContext plain and simple.

RZR-UA (approved):

Issue #1622, opened 2026-04-10
IndexedDB cache not force-refreshing when data changes - Favorites groups appear empty 


Rowin63:
<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/f8ecad02-e994-4e2e-8c9c-ea23e2b84a51"  width="300" height="auto"/> </td>
  </tr>
</table>

Rowin63:
This seems to be a Firefox issue again, on a Windows 11 Pro 25H2. 
But no warnings, no errors appear in the "tools / web developer tools". It´s empty.

Rowin63:
Solved it the same way as every time before: had to delete the "indexed database" again, then Osmand Web works again.

I want to know:

Is there anyting I can do on MY side to prevent this situations?
Is it a Firefox-issue?
is it an Osmand-Web-issue?

It´s just annoying, gettig into this non-working-situations again and again.

![Image](https://github.com/user-attachments/assets/e9c14d0d-6280-4550-b743-ea7fb9928e1c)

alisa911:
@Rowin63 Yes, this is an issue on our side. The mechanism for force-refreshing the browser's IndexedDB cache when data changes is not fully implemented. Specifically, the cached favorite groups can become stale or incomplete, and the app doesn't currently detect or recover from this state automatically, which is why manually deleting the Indexed Database resolves it every time.
We'll work on fixing this in the near future. Thank you for reporting and for the detailed description, it helps a lot!

PR #1634, opened 2026-04-15
add versions for indexdb

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1622

map/src/index.js
```diff
@@ -3,10 +3,16 @@ import './assets/css/index.css';
...
+import { ensureAppDataVersion } from './util/appDataVersion';
 
 const container = document.getElementById('root');
 const root = createRoot(container);
 
+try {
+    await ensureAppDataVersion();
+} catch (e) {
+    console.error(e);
+}
```
RZR-UA, line 15: This isn't the best place for the ensureAppDataVersion() call.

RZR-UA (approved):

map/src/index.js
alisa911: This is the right place. Storage migration has to finish before the tree mounts so nothing opens IndexedDB mid-migration. Awaiting here before render is the simplest way to guarantee that order.

PR #1639, opened 2026-04-16
fix empty name in local tracks

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1616

Issue #543, opened 2024-10-15
Redesign Add favorite UI


dmpr0:
## ToDo
> Related
> * #677

## Add / Edit favorite
<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/179e6638-2b16-4638-9d4e-d0b078796380"  width="300" height="auto"/> </td>
  </tr>
</table>   

## Map

- [x] Show icon preview with in hover state with big pin, should show selected shape, color, icon

## Name

- [x] Use the same **TextField** component that is already used on the **Login** page.
- [ ] Max 3 lines of text, then enable vertical scroll.

## Address, [Figma](https://www.figma.com/design/fqiiJCrepmJP7XkA3YixSq/My-Places?node-id=192-35319&m=dev)

**State 1:**
<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/a2a7e229-8faa-486d-8e27-2d68ff6ae623"  width="300" height="auto"/> </td>
  </tr>
</table>  

During automatic address search, right after adding, or when the user clears the text field and clicks the Location icon.

- [x] Label – `Address`
- [x] Primary value – `Searching...`
- [x] Trailing action – Clear field, icon – `ic_action_cancel`

**State 2:**

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/1ff1cf61-4e24-43d8-9d0f-6fd1c6c7935c"  width="300" height="auto"/> </td>
  </tr>
</table>  

After the user clears the text field.

- [x] Label – `Address`
- [x] Trailing action – `Auto-fill address`, icon – `ic_action_location_marker_outlined`
- - [x] Add tooltip on hover

**State 3:**

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/68e43299-115d-45dd-af78-b59203b8493d"  width="300" height="auto"/> </td>
  </tr>
</table>  

Filled text field with Address.

- [x] Label – `Address`
- [x] Primary value - typed or automatically determined address.
- [x] Trailing action – Clear field, icon – `ic_action_cancel`

### Empty state:
<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/8b015db7-55c5-48a8-9a3d-5214ea2c4c62"  width="300" height="auto"/> </td>
  </tr>
</table>   

- [x] Clicking on "Add notes" or the whole item should open the secondary panel.

### Filled state:
[...]

PR #1695, opened 2026-05-07
Fix stale preview marker remaining on map after WptEditPanel closes

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/543

PR #1703, opened 2026-05-11
Fix favorites group commit and new-group save

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/543

Issue #1557, opened 2026-03-24
Not posible to scroll all the way down


Rowin63:
Again, it´s not possible to scroll all the way down in this track list.
Firefox, Chrome.

![Image](https://github.com/user-attachments/assets/f4bccdd6-0d0f-4e45-9019-31fac657dcb5)

dmpr0:
## ToDo

- [ ] Limit AppBar to one line

Zirochkabila:
# Testing 1

> Web - prod

The list cannot be scrolled all the way down. The bottom items are partially hidden or inaccessible because the multi-line AppBar/header takes too much vertical space - **Fixed** - The folder name is limited to one line



<details><summary>Details</summary>
<video src="https://github.com/user-attachments/assets/960cd96e-5ff7-4fed-bcd7-f302775293a0"></video>
</details>

PR #1707, opened 2026-05-12
Fix limit AppBar title to one line in tracks and favorites menu

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1557

PR #1708, opened 2026-05-12
Fix search pin and icon flickering on zoom

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1600

Issue #1719, opened 2026-05-18
Track edits fail to save to the Cloud


dmpr0:
## Description
When modifying a previously saved track and attempting to sync the updates, the edits fail to save to the **Cloud**.

### How to reproduce
* Create a new route using **Plan a Route**, then tap **Save**.
* Open the track and tap **Edit**.
* Make some changes, then tap **Save to Cloud**.

### Actual result

<table style="width:100%">
  <tr>
    <td><video src="https://github.com/user-attachments/assets/42d203c4-64d0-4416-a7ea-ece6a6e1f59c"  width="300" height="auto"/> </td>
  </tr>
</table>  

* The changes do not save to the **Cloud**.

### Expected result
* All user changes should be successfully saved to the **Cloud**.

**Environment**
OsmAnd Version: 1.03

Zirochkabila:
# Testing 1

> Web - prod

The changes do not save to the Cloud - **Fixed**

<details><summary>Details</summary>
<video src="https://github.com/user-attachments/assets/1c8a666f-65c4-423c-9b45-a83d6f73e913"></video>
</details>

PR #1721, opened 2026-05-20
Fix Blob body not hashed in API cache key

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1719

map/src/manager/track/TracksManager.js
```diff
@@ -483,11 +483,13 @@ export function addDistanceToPoints(points) {
 export async function getGpxFileFromTrackData(file, routeTypes, simplified = false) {
     const trackData = prepareTrackData({ file, routeTypes });
     const compressedData = compressJSONToBlob(trackData);
+    const apiCacheKey = await digest(JSON.stringify({ trackData, simplified }));
```
RZR-UA: JSON.stringify() is expensive; please avoid it.

map/src/manager/track/TracksManager.js
```diff
@@ -483,11 +483,13 @@ export function addDistanceToPoints(points) {
...
     return await apiPost(`${process.env.REACT_APP_GPX_API}/gpx/save-track-data`, compressedData, {
         params: {
             simplified,
         },
         apiCache: true,
+        apiCacheKey,
```
RZR-UA: External apiCacheKey breaks library abstraction. Better to support Blob in apiPost() or generateCacheKey().

RZR-UA (changes requested):

RZR-UA (approved):

PR #1745, opened 2026-05-28
Add live location sharing

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/3153

map/src/context/LiveTrackingContext.js
```diff
@@ -0,0 +1,113 @@
...
+    const [selectedLiveTranslation, setSelectedLiveTranslation] = useState(null);
+    const [followLiveLocation, setFollowLiveLocation] = useState(null);
+    const [myBroadcastTid, setMyBroadcastTid] = useState(null);
+    const [isMyBroadcastPaused, setIsMyBroadcastPaused] = useState(false);
+
+    const livePath = MAIN_URL_WITH_SLASH + LIVE_TRACKS_URL;
+    const openLiveTracks = location.pathname.startsWith(livePath);
+
+    // Sync the selected translation from the URL (?tid=...#<key>), restoring the AES key the map
+    // stashed in sessionStorage before leaflet-hash cleared the fragment.
```
RZR-UA, line 40: Think about implementing a URL scheme with a hash saved in the query string. For example, `/map/live/?hash=17/50/40&tid=xxx#key` may store/clear the `key` and finally restore the leaflet `hash` to keep saved map zoom/lat/lon.

map/src/util/hooks/live/useLiveTracking.js
```diff
@@ -0,0 +1,740 @@
...
+                    encryptLocation(key, locationData)
+                        .then((encData) => {
+                            apiPost(
+                                '/mapapi/translation/msg',
+                                `translationId=${encodeURIComponent(tid)}&encryptedData=${encodeURIComponent(encData)}`,
+                                { headers: { 'Content-Type': 'application/x-www-form-urlencoded' } }
+                            ).catch(() => {});
+                        })
+                        .catch(() => {});
+                }
```
RZR-UA: 1. Add throttling (limit not to send POST queries too often, at least 1 per second)
2. Add reaction on HTTP errors (at least, console.info, at most increase throttling on errors)
3. The same points should be considered in Android code (which is even more important to avoid server overloads).

RZR-UA (approved):

PR #1773, opened 2026-06-04
Filter invisible tags

Dima-1:
https://github.com/osmandapp/OsmAnd/issues/24809


alisa911 (commented):
We need a solution to ensure that filtered tags on the web match those on Android. The final list of tags should always be consistent for Favorites and POI.

vshcherb:
Definitely not a hard code that's totally wrong.
Possible solution
1. Server needs to process tags with translations 
2. Server needs to provide a list / map to exclude with translations and some logic

This needs to be maintained in sync with Android

Issue #1283, opened 2025-12-04
Add possibility to import tracks (.gpx files) dragging and dropping them onto the map


DmitryAlexei:
### Description
As a user, I would like a simpler and more convenient way to import tracks and instantly displaying them on the map with the context menu open. That would be especially useful when the needed track isn't labeled correctly (e.g., only containing numbers or a date in its name).

### Actual situation
Currently Web map doesn't support importing tracks by dragging and dropping them onto the map. 

### Proposed solution
Something like gpx.studio and others.

dmpr0:
## ToDo

- [x] Upload files to the OsmAnd Cloud, not local storage

- - - - - -

 <table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/a5785a60-377c-44cf-bcf3-036a94c582a0"  width="500" height="auto"/> </td>
  </tr>
</table>   

- [x] Dragging and dropping into the map should import the track into the Tracks / **Import** folder
- [x] Highlight only the visible map area; don't draw the area below the left panel

- - - - - -
 
<table style="width:100%">
  <tr>
    <th>Open Folder</th>
    <th>Folder item</th>
    <th>Area</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/cc99c87d-124d-4390-9f77-90122e796163"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/d1359f31-26a1-4494-83c8-355c52835554"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/dfef62ed-4fa1-483e-a346-01fb6ce5c673"  width="300" height="auto"/> </td>
  </tr>
</table>   

- [x] Hovering over a folder should highlight this folder and import the file into it
- [x] Hovering over tracks in any folder should highlight this area and import the file into this folder
- [x] Show the imported file item, just as it appears after importing through the file picker

Dima-1:
<table style="width:100%">
  <tr>
    <th>Map</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/24ea2c0e-e1a8-407c-a499-7376882ca2c6"  width="500" height="auto"/> </td>
  </tr>
</table> 
<table style="width:100%">
  <tr>
    <th>Open Folder</th>
    <th>Folder item</th>
    <th>Area</th>
  </tr>
  <tr>
[...]

PR #1799, opened 2026-06-16
Improved drag and drop

Dima-1:
https://github.com/osmandapp/web/issues/1283


map/src/frame/TracksFileDragController.jsx
```diff
@@ -0,0 +1,101 @@
...
+        const onDragOver = (e) => {
+            if (!hasFiles(e)) {
+                return;
+            }
+            e.preventDefault();
+            e.dataTransfer.dropEffect = 'copy';
+
+            const currentCtx = ctxRef.current;
+            const { hoverFolder, overMap } = resolveGpxDropTarget(e.clientX, e.clientY, currentCtx);
+            currentCtx.setGpxFileDrag({ active: true, hoverFolder, overMap });
```
alisa911: setGpxFileDrag is called with a fresh object on every dragover, spamming context updates and re-rendering on each mouse move

map/src/frame/TracksFileDragController.jsx
```diff
@@ -0,0 +1,101 @@
...
+        };
```
alisa911: the overlay is activated regardless of auth, so free/anonymous users see "Drop GPX to import" but the drop silently no-ops

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,99 @@
...
+                return;
+            }
+
+            const files = Array.from(fileList || []).filter(isCloudTrackFile);
+            if (files.length === 0) {
+                return;
+            }
+
+            const selected = files.length === 1;
+            ctx.setTrackLoading(files.map((track) => removeFileExtension(track.name) + GPX_FILE_EXT));
```
alisa911: loading entries are set as name + GPX_FILE_EXT but removed by raw file.name on error, so failed .kmz/.kml imports leave a stuck loading placeholder

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,99 @@
+import { useCallback, useContext, useEffect } from 'react';
+import AppContext from '../../context/AppContext';
+import LoginContext from '../../context/LoginContext';
+import { FREE_ACCOUNT } from '../../manager/LoginManager';
+import { GPX_FILE_EXT, KMZ_FILE_EXT } from '../../manager/track/TracksManager';
+import { createTrackFreeName, removeFileExtension, saveTrackToCloud } from '../../manager/track/SaveTrackManager';
+import { useMutator } from '../Utils';
+
+const CLOUD_TRACK_EXTENSIONS = ['.gpx', '.kmz', '.kml'];
```
alisa911: .kml is accepted but always saved with type: 'GPX' and read as text

map/src/manager/track/TracksManager.js
```diff
@@ -57,8 +57,8 @@ const HOURS_24_MS = 86400000;
 const AUTO_SRTM_MAX_POINTS = 10000; // don't overload Auto-SRTM API with huge OSRM tracks
 const AUTO_SRTM_MIN_BAD_POINTS_PERCENT = 10; // limit by % of no-elevation points (type=osmand might have up to 10%)
 export const FIT_BOUNDS_MAX_ZOOM = 17;
-export const FIT_BOUNDS_EXTRA_PADDING = 10;
 export const DEFAULT_GROUP_NAME = '';
+export const IMPORT_FOLDER_NAME = 'Import';
```
alisa911, line 62: IMPORT_FOLDER_NAME = 'Import' is a hardcoded, untranslated cloud-folder name; verify it matches existing folder-naming conventions

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,99 @@
...
+
+function isCloudTrackFile(file) {
+    const name = file?.name?.toLowerCase() ?? '';
+
+    return CLOUD_TRACK_EXTENSIONS.some((ext) => name.endsWith(ext));
+}
+
+function validName(name) {
```
alisa911: validName is redefined locally here but an exported validName already exists in TracksManager.js

map/src/frame/TracksMapDropGeometry.js
```diff
@@ -0,0 +1,159 @@
+import { HEADER_SIZE, MAIN_MENU_MIN_SIZE, MAIN_MENU_OPEN_SIZE } from '../../manager/GlobalManager';
+
+export const OVERLAY_MARGIN = 16;
+
+export function getVisibleMapInsets(ctx) {
```
alisa911: Visible-map-area geometry already exists: MapStateLayer.calcVisibleBboxParamsPx / getVisibleBboxInfo and mtx.fitBoundsPadding. This reimplements the same left/top/bottom insets from the same constants, and the two already diverge (this uses MAIN_MENU_OPEN_SIZE on open menu while MapStateLayer always uses MAIN_MENU_MIN_SIZE). Reuse the existing source of truth instead of a parallel computation.

map/src/context/AppContext.js
```diff
@@ -135,6 +127,8 @@ export const AppContextProvider = (props) => {
     const [globalConfirmation, setGlobalConfirmation] = useState(null);
 
     const [openMenu, setOpenMenu] = useState(null);
+    const [openMainMenu, setOpenMainMenu] = useState(false);
```
alisa911: openMainMenu was lifted from GlobalFrame local state into global AppContext solely so GpxMapDropGeometry can read it. But the canonical visible-area logic (MapStateLayer.calcVisibleBboxParamsPx) intentionally ignores menu-open state (the expanded menu is an overlay, not a layout shift), so the overlay shouldn't need it either. Drop the parallel geometry, reuse the existing visible-bounds, and keep openMainMenu local.

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,99 @@
...
+                }).then();
+                mutateUploadedFiles((o) => delete o[file]);
+                break; // process 1 file per 1 render
+            }
+        }
+    }, [uploadedFiles]);
+
+    const importGpxFiles = useCallback(
+        (fileList, folder) => {
+            if (!ltx.loginUser || ltx.accountInfo?.account === FREE_ACCOUNT) {
```
alisa911: Re-implements an auth check that already exists

map/src/frame/TracksMapDropGeometry.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+    const folderEl = el.closest('[data-cloud-track-folder]');
+    if (!folderEl) {
+        return null;
+    }
+
+    return folderEl.getAttribute('data-cloud-track-folder') ?? '';
+}
+
+export function resolveGpxDropTarget(clientX, clientY, ctx) {
```
alisa911: Determining the hovered folder by scraping the DOM (elementFromPoint, getElementById, data-*, id-prefix selectors) duplicates state the React tree already owns. Set the hover target from onDragEnter/onDragOver handlers on the folder components into context instead of reading layout back out of the DOM.

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,99 @@
+import { useCallback, useContext, useEffect } from 'react';
+import AppContext from '../../context/AppContext';
+import LoginContext from '../../context/LoginContext';
+import { FREE_ACCOUNT } from '../../manager/LoginManager';
+import { GPX_FILE_EXT, KMZ_FILE_EXT } from '../../manager/track/TracksManager';
+import { createTrackFreeName, removeFileExtension, saveTrackToCloud } from '../../manager/track/SaveTrackManager';
+import { useMutator } from '../Utils';
+
+const CLOUD_TRACK_EXTENSIONS = ['.gpx', '.kmz', '.kml'];
+
```
alisa911, line 11: useCloudGpxImport.js vs useLocalGpxImport.js — These two hooks share the same scaffold almost verbatim: useMutator({}), the one-file-per-render useEffect with break, and the Array.from(fileList).filter(...) + FileReader flow (including the // process 1 file per 1 render comment). Extract the shared import skeleton and parameterize the save target (local vs cloud) instead of maintaining two near-identical copies.

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,99 @@
...
+                                    originalName: file.name,
+                                    folder,
+                                })
+                        );
+                    } else {
+                        ctx.setTrackErrorMsg({
+                            title: 'Import error',
+                            msg: `Unable to import ${file.name}`,
+                        });
+                        ctx.setTrackLoading([...ctx.trackLoading.filter((n) => n !== file.name)]);
```
alisa911: Inside the async FileReader load callback, ctx.trackLoading is a stale closure captured when the reader was created; with multiple files in flight the filter operates on an outdated array and drops the wrong entries. Use the functional updater: setTrackLoading(prev => prev.filter((n) => n !== file.name)).

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,99 @@
...
+
+    useEffect(() => {
+        for (const file in uploadedFiles) {
+            const uploadedFile = uploadedFiles[file];
+            let open = uploadedFile.selected;
+            let fileName = uploadedFile.name;
+            const folder = uploadedFile.folder;
+            if (validName(fileName)) {
+                fileName = removeFileExtension(fileName);
+                fileName = createTrackFreeName(fileName, ctx.tracksGroups, folder);
```
alisa911: Multi-file import is racy: saveTrackToCloud(...).then() is fire-and-forget, so the effect deletes the entry and moves to the next file before the save finishes and ctx.tracksGroups updates. As a result createTrackFreeName runs against a stale group list and several files resolving to the same base name can collide/overwrite. Await each save (or compute unique names up front) before processing the next file.

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,99 @@
...
+            ctx.setTrackLoading(files.map((track) => removeFileExtension(track.name) + GPX_FILE_EXT));
+
+            files.forEach((file) => {
+                const reader = new FileReader();
+                reader.addEventListener('load', (e) => {
+                    const data = e.target.result;
+                    if (data) {
+                        mutateUploadedFiles(
+                            (o) =>
+                                (o[file.name] = {
```
alisa911: uploadedFiles is keyed by file.name, so two dropped files with the same name silently overwrite each other; key by a unique id instead.

map/src/frame/TracksFileDragController.jsx
```diff
@@ -0,0 +1,101 @@
...
+                importGpxFiles(files, IMPORT_FOLDER_NAME);
+            }
+            resetDrag();
+        };
+
+        const onDragEnd = () => {
+            resetDrag();
+        };
+
+        window.addEventListener('dragenter', onDragEnter);
```
alisa911: Listeners are attached to window with a global preventDefault on dragover, intercepting every drag&drop in the app and overriding default behavior page-wide. This can clash with other drop zones / file inputs (e.g. favorites import). Scope the listeners to the app container, or confirm no other consumer relies on native drop.

map/src/frame/TracksFileDragController.jsx
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export default function GpxFileDragController() {
+    const ctx = useContext(AppContext);
+    const { importGpxFiles } = useCloudGpxImport();
+    const dragCounterRef = useRef(0);
+    const ctxRef = useRef(ctx);
+    ctxRef.current = ctx;
+
+    useEffect(() => {
```
alisa911, line 15: The effect depends on importGpxFiles, whose identity changes on nearly every render (its useCallback depends on the whole ctx), so all five window listeners are torn down and re-added constantly. This also defeats the ctxRef indirection that exists to keep them stable. Keep importGpxFiles in a ref too (or make it stable) so the listeners are bound once.

map/src/frame/TracksFileDragController.jsx
```diff
@@ -0,0 +1,101 @@
...
+function hasFiles(e) {
+    return e.dataTransfer?.types?.includes('Files');
+}
+
+export default function GpxFileDragController() {
+    const ctx = useContext(AppContext);
+    const { importGpxFiles } = useCloudGpxImport();
+    const dragCounterRef = useRef(0);
+    const ctxRef = useRef(ctx);
+    ctxRef.current = ctx;
```
alisa911: Reading the latest ctx via a ref is only justified if the listeners are bound once, but they aren't, so this is currently the worst of both: re-subscription and a ref. Also ctxRef.current = ctx mutates a ref during render. Either make the listeners truly stable and keep the ref, or drop the ref and depend on ctx directly.

Use CLAUDE md!

map/src/map/components/GpxMenuDropOverlay.jsx
```diff
@@ -0,0 +1,68 @@
...
+import { useContext, useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import AppContext from '../../context/AppContext';
+import { getMenuDropOverlayRect, getMenuOverlayContainer } from './GpxMapDropGeometry';
+import styles from './gpxMapDropOverlay.module.css';
+
+export default function GpxMenuDropOverlay() {
+    const ctx = useContext(AppContext);
+    const { t } = useTranslation();
+    const ctxRef = useRef(ctx);
```
alisa911: const ctxRef = useRef(ctx) violates the project rule "refs are a last resort" (CLAUDE.md), and stashing the entire context in a ref is exactly what we don't do.

map/src/map/components/GpxMenuDropOverlay.jsx
```diff
@@ -0,0 +1,68 @@
+import { Typography } from '@mui/material';
+import { useContext, useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import AppContext from '../../context/AppContext';
+import { getMenuDropOverlayRect, getMenuOverlayContainer } from './GpxMapDropGeometry';
+import styles from './gpxMapDropOverlay.module.css';
+
+export default function GpxMenuDropOverlay() {
```
alisa911: GpxMenuDropOverlay.jsx (return) / GpxMapDropOverlay.jsx — Both overlays render identical markup, the same web:drop_gpx_to_import label and the same positioning approach. Extract a shared presentational DropOverlay component that just takes a rect/insets, instead of duplicating the JSX in two places.

map/src/map/components/GpxMenuDropOverlay.jsx
```diff
@@ -0,0 +1,68 @@
...
+    const ctx = useContext(AppContext);
+    const { t } = useTranslation();
+    const ctxRef = useRef(ctx);
+    const [overlayStyle, setOverlayStyle] = useState(null);
```
alisa911: The hovered folder is already highlighted via the groupDropTarget/folderDropTarget CSS classes; a separately measured portal overlay on top of the same folder duplicates that visual feedback. Pick one mechanism.

map/src/map/components/GpxMenuDropOverlay.jsx
```diff
@@ -0,0 +1,68 @@
...
+    const hoverFolder = ctx.gpxFileDrag?.hoverFolder;
+    const active = ctx.gpxFileDrag?.active && !ctx.gpxFileDrag?.overMap && hoverFolder !== null;
+
+    useEffect(() => {
+        if (!active) {
+            setOverlayStyle(null);
+            return;
+        }
+
+        const update = () => {
```
alisa911: Locating and measuring the target folder via getElementById to position a fixed overlay couples this to DOM ids and is brittle; prefer rendering the highlight inside the folder component itself.

map/src/menu/tracks/TrackGroupFolder.jsx
```diff
@@ -138,10 +139,14 @@ export default function TrackGroupFolder({ folder = null, smartf = null }) {
...
     }
 
+    const isDropTarget = group && group.type !== SMART_TYPE && !smartf;
+
     return (
         <>
             <Box
                 id={`se-tracks-folder`}
+                className={isDropTarget ? styles.folderDropTarget : undefined}
+                data-cloud-track-folder={isDropTarget ? group.fullName : undefined}
```
alisa911: data-cloud-track-folder exists only to bridge the global window drag listener (which has just cursor coords) back to a folder identity via elementFromPoint(...).closest(...). This is DOM-scraping for state React already owns. Attach onDragEnter/onDragOver/onDrop to the folder components (they already close over group.fullName) and set the hover target in context, removing the data-* stamping entirely. If kept short-term, at least extract the attribute name into a shared constant and document the ""-means-root sentinel.

alisa911 (commented):
Rename all the new files away from the gpx prefix, the feature handles not only GPX (also KMZ/KML), but only tracks (not favorites). Also move these files out of map/ into frame/, since they're not map-specific.

map/src/manager/track/TracksManager.js
Dima-1: Localization for the import folder name is not needed

map/src/util/hooks/useCloudGpxImport.js
Dima-1: How is it possible to drag and drop two files with the same name?

map/src/map/components/GpxMenuDropOverlay.jsx
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/507eeee27ac2574d5e2ca912ffe33bc3e15ee7ff

map/src/frame/TracksFileDragController.jsx
Dima-1: https://github.com/osmandapp/web/commit/507eeee27ac2574d5e2ca912ffe33bc3e15ee7ff

map/src/util/hooks/useCloudGpxImport.js
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/458f7939078e0c2cc8dc797ae3276ab128450a75

map/src/util/hooks/useCloudGpxImport.js
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/507eeee27ac2574d5e2ca912ffe33bc3e15ee7ff

map/src/util/hooks/useCloudGpxImport.js
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/458f7939078e0c2cc8dc797ae3276ab128450a75

map/src/frame/TracksFileDragController.jsx
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/55f689aa04bacfa565ab2c94dddda820da1cba70

map/src/frame/TracksFileDragController.jsx
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/458f7939078e0c2cc8dc797ae3276ab128450a75

map/src/map/components/GpxMenuDropOverlay.jsx
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/55f689aa04bacfa565ab2c94dddda820da1cba70

map/src/map/components/GpxMenuDropOverlay.jsx
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/507eeee27ac2574d5e2ca912ffe33bc3e15ee7ff

map/src/util/hooks/useCloudGpxImport.js
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/458f7939078e0c2cc8dc797ae3276ab128450a75

map/src/frame/TracksFileDragController.jsx
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/55f689aa04bacfa565ab2c94dddda820da1cba70

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,99 @@
...
+                                })
+                        );
+                    } else {
+                        ctx.setTrackErrorMsg({
+                            title: 'Import error',
+                            msg: `Unable to import ${file.name}`,
+                        });
+                        ctx.setTrackLoading([...ctx.trackLoading.filter((n) => n !== file.name)]);
+                    }
+                });
```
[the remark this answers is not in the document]
Dima-1: https://github.com/osmandapp/web/commit/458f7939078e0c2cc8dc797ae3276ab128450a75

map/src/frame/TracksMapDropGeometry.js
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/55f689aa04bacfa565ab2c94dddda820da1cba70

map/src/frame/TracksFileDragController.jsx
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/458f7939078e0c2cc8dc797ae3276ab128450a75

map/src/menu/tracks/TrackGroupFolder.jsx
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/55f689aa04bacfa565ab2c94dddda820da1cba70

map/src/map/components/GpxMenuDropOverlay.jsx
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/9e9bc4ab4d7030dafc4edb584d128429c6385d3f

map/src/util/hooks/useCloudGpxImport.js
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/1d721f83e642c69f230a7c8bc37474950c589535

map/src/context/AppContext.js
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/2bcd8523bee981385bb78d524c5ffc98506ac9f7

map/src/frame/TracksMapDropGeometry.js
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/2bcd8523bee981385bb78d524c5ffc98506ac9f7

map/src/util/hooks/useCloudGpxImport.js
Dima-1: KML is imported as GPX

map/src/frame/TracksFileDragController.jsx
```diff
@@ -0,0 +1,94 @@
...
+    return e.dataTransfer?.types?.includes('Files');
+}
+
+export default function TracksFileDragController() {
+    const ctx = useContext(AppContext);
+    const ltx = useContext(LoginContext);
+    const { importGpxFiles } = useCloudGpxImport();
+    const isProAccount = ltx.isProAccount();
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/39ea0ee63582f963816009ca7d6b1323cb27957d

map/src/frame/TracksFileDragController.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+        const onDrop = (e) => {
+            if (!hasFiles(e) || !isProAccount) {
+                return;
+            }
+            e.preventDefault();
+
+            const files = Array.from(e.dataTransfer?.files || []);
+            const { hoverFolder, overMap } = ctx.gpxFileDrag ?? GPX_FILE_DRAG_IDLE;
+            if (hoverFolder !== null) {
```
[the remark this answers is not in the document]
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/39ea0ee63582f963816009ca7d6b1323cb27957d

map/src/frame/TracksFileDragController.jsx
```diff
@@ -0,0 +1,94 @@
...
+        // dragend has no preventDefault; keep on window so external file drags still reset state.
+        window.addEventListener('dragend', onDragEnd);
+        return () => {
+            container.removeEventListener('dragenter', onDragEnter);
+            container.removeEventListener('dragover', onDragOver);
+            container.removeEventListener('dragleave', onDragLeave);
+            container.removeEventListener('drop', onDrop);
+            window.removeEventListener('dragend', onDragEnd);
+        };
+    }, [ctx.gpxFileDrag, ctx.setGpxFileDrag, importGpxFiles, isProAccount]);
```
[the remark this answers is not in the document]
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/39ea0ee63582f963816009ca7d6b1323cb27957d

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,94 @@
...
+                    });
+                } else {
+                    ctx.setTrackErrorMsg({
+                        title: 'Import error',
+                        msg: `Unable to import ${file.name}`,
+                    });
+                    ctx.setTrackLoading((prev) => prev.filter((n) => n !== loadingName));
+                    resolve(null);
+                }
+            });
```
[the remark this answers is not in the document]
Dima-1: https://github.com/osmandapp/web/pull/1799/commits/f3b767edc15a19dbc6778dde3509b2a18df1e430

map/src/frame/TracksFileDragController.jsx
```diff
@@ -0,0 +1,93 @@
...
+        if (!container) {
+            return;
+        }
+
+        const resetDrag = () => {
+            ctx.setGpxFileDrag(GPX_FILE_DRAG_IDLE);
+        };
+
+        const onDragEnter = (e) => {
+            if (!hasFiles(e) || !ltx.isProAccount()) {
```
RZR-UA: Why is it only ProAccount? (3x times) Don't we allow 5 MB of cloud storage for free accounts?

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,116 @@
...
+export default function useCloudGpxImport() {
+    const ctx = useContext(AppContext);
+    const ltx = useContext(LoginContext);
+    const { t } = useTranslation();
+
+    const canImport = useCallback(() => ltx.isProAccount(), [ltx]);
+
+    const onFilesSelected = useCallback(
+        (files, folder) => {
+            const batchTaken = new Set();
```
RZR-UA: occupiedFileNames instead of batchTaken

map/src/manager/track/SaveTrackManager.js
```diff
@@ -162,7 +161,7 @@ export function removeFileExtension(filename) {
     return filename.includes('.') ? filename.slice(0, filename.lastIndexOf('.')) : filename;
 }
 
-export function createTrackFreeName(name, otherTracks, folder = null, folderName = null) {
+export function createTrackFreeName(name, otherTracks, folder = null, folderName = null, extraOccupied = null) {
```
RZR-UA: What is extra? Who is occupied? Use `occupiedFileNames`

map/src/manager/track/SaveTrackManager.js
```diff
@@ -441,6 +440,9 @@ async function downloadAfterUpload(ctx, file, showOnMap) {
         });
         newGpxFiles[file.name].analysis = TracksManager.prepareAnalysis(newGpxFiles[file.name].analysis);
         newGpxFiles[file.name].showOnMap = showOnMap;
+        if (showOnMap) {
+            newGpxFiles[file.name].zoomToTrack = true;
+        }
```
RZR-UA: Why? We usually avoid any map zoom/position change without such user-initiated action.

map/src/manager/track/TracksManager.js
```diff
@@ -59,6 +59,7 @@ const AUTO_SRTM_MIN_BAD_POINTS_PERCENT = 10; // limit by % of no-elevation point
 export const FIT_BOUNDS_MAX_ZOOM = 17;
 export const FIT_BOUNDS_EXTRA_PADDING = 10;
 export const DEFAULT_GROUP_NAME = '';
+export const IMPORT_FOLDER_NAME = 'Import';
```
RZR-UA, line 62: I don't think we need to import into the "Import" folder. In Android, we import tracks into the current folder (including the main root folder). It would be better to use the same approach: when a user drops a file into the map area, we import it directly into the current folder, even into the root folder.

map/src/menu/tracks/TrackLoading.jsx
```diff
@@ -11,7 +11,7 @@ export default function TrackLoading({ name }) {
             </ListItemIcon>
             <ListItemText>
                 <Typography className={styles.groupName} noWrap>
-                    {name}
+                    {name.replace(/\.[^.]+$/, '')}
```
RZR-UA: Why is part of the track name cut? Weird regexp...

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,116 @@
+import { useCallback, useContext } from 'react';
+import { useTranslation } from 'react-i18next';
+import AppContext from '../../context/AppContext';
+import LoginContext from '../../context/LoginContext';
+import { GPX_FILE_EXT, KMZ_FILE_EXT, validName } from '../../manager/track/TracksManager';
+import { createTrackFreeName, removeFileExtension, saveTrackToCloud } from '../../manager/track/SaveTrackManager';
+import useGpxImport from './useGpxImport';
+
+const CLOUD_TRACK_EXTENSIONS = ['.gpx', '.kmz', '.kml'];
```
RZR-UA: Use this CLOUD_TRACK_EXTENSIONS to make `accept=".gpx, .kmz, .kml"` strings in the codebase.

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,116 @@
...
+
+export default function useCloudGpxImport() {
+    const ctx = useContext(AppContext);
+    const ltx = useContext(LoginContext);
+    const { t } = useTranslation();
+
+    const canImport = useCallback(() => ltx.isProAccount(), [ltx]);
```
RZR-UA, line 17: isProAccount seems incorrect

map/src/util/hooks/useCloudGpxImport.js
```diff
@@ -0,0 +1,116 @@
...
+                        onFail();
+                    }
+                });
+                reader.addEventListener('error', onFail);
+                reader.addEventListener('abort', onFail);
+                if (file.name.toLowerCase().endsWith(KMZ_FILE_EXT)) {
+                    reader.readAsArrayBuffer(file);
+                } else {
+                    reader.readAsText(file);
+                }
```
RZR-UA, line 80: Please test carefully on each of the GPX, KML, and KMZ files.

map/src/util/hooks/useGpxImport.js
```diff
@@ -0,0 +1,40 @@
+import { useCallback, useEffect } from 'react';
+import { useMutator } from '../Utils';
+
+export default function useGpxImport({ canImport, isTrackFile, onFilesSelected, readFile, saveFile }) {
+    const [uploadedFiles, mutateUploadedFiles] = useMutator({});
+
+    const importGpxFiles = useCallback(
+        (fileList, folder) => {
+            if (canImport && !canImport()) {
```
RZR-UA: Please, never mix a variable and a method with the same name.

map/src/util/hooks/useGpxImport.js
```diff
@@ -0,0 +1,40 @@
...
+                return;
+            }
+
+            const files = Array.from(fileList || []).filter(isTrackFile);
```
RZR-UA, line 13: Weird filter which does not depend on array elements...

map/src/util/hooks/useLocalGpxImport.js
```diff
@@ -1,41 +1,41 @@
-import { useCallback, useContext, useEffect } from 'react';
+import { useCallback, useContext } from 'react';
 import AppContext from '../../context/AppContext';
 import TracksManager from '../../manager/track/TracksManager';
 import { saveTrackToLocal } from '../../manager/track/SaveTrackManager';
-import { useMutator } from '../Utils';
+import useGpxImport from './useGpxImport';
+
+const LOCAL_TRACK_EXTENSION = '.gpx';
```
RZR-UA: We already have GPX_FILE_EXT. Remove redundant '.gpx' or the dupe const itself.

map/src/util/hooks/useLocalGpxImport.js
```diff
@@ -1,41 +1,41 @@
...
+    return useGpxImport({
+        isTrackFile: isLocalTrackFile,
+        readFile,
+        saveFile,
+    });
+}
+
+function isLocalTrackFile(file) {
+    return file?.name?.toLowerCase().endsWith(LOCAL_TRACK_EXTENSION);
 }
```
RZR-UA, line 39: Why have you modified this file (**local** track) if the Issue is about drag-and-drop import into **cloud** tracks?

alisa911 (commented):
1. Event model relies on implicit listener order. TracksFileDragController attaches native listeners on #root, while the drop zones use React synthetic handlers delegated to the same #root. The guard's e.stopPropagation() is synthetic — it calls nativeEvent.stopPropagation() but not stopImmediatePropagation(), so it does not stop the controller's native listener on the same node. It only works today because React registers its root listener before the controller's useEffect runs. Move everything to a single layer: React handlers on the zones plus one onDrop, and drop the global native listener.
2. Remove gpxFileDragRef. It only exists because the single global onDrop reads stale closure state. Give each zone its own onDrop (they already own onDragEnter/onDragOver) so the folder knows its fullName directly. Remove the global native drop listener at the same time — otherwise synthetic stopPropagation won't stop it and you'll get a double import.
3. Simplify TracksDropHighlight and drop the extra refs. getItemOffset (added to useImperativeHandle) plus outerRef exist only to offset the highlight past folder rows via getBoundingClientRect/scroll-sync, and it's O(n) on every scroll/resize. Render the highlight inside the scroll container (absolute relative to content) instead of measuring it externally — then scrollRef, listRef, getItemOffset and outerRef all go away, leaving a single container ref.
4. Fix the name race on concurrent drops. onFilesSelected does ctx.setTrackLoading(freeNames.map(...)) (full replace, not append), and batchTaken in createTrackFreeName only dedupes within a single drop while tracksGroups isn't updated until upload finishes. Two quick drops of a same-named file into one folder both resolve to the same free name → cloud collision, and the second drop wipes the first's loading indicators. Append to trackLoading and account for in-flight names.

Order I'd suggest: (1) event model → (2) remove gpxFileDragRef → (3) simplify the highlight.

map/src/manager/track/TracksManager.js
Dima-1: Dragging and dropping into the map should import the track into the Tracks / Import folder

map/src/frame/TracksFileDragController.jsx
Dima-1: The free 5 MB cloud storage is used only for favorites and settings. Track storage is available to Pro users only

map/src/util/hooks/useCloudGpxImport.js
Dima-1: The free 5 MB cloud storage is used only for favorites and settings. Track storage is available to Pro users only

map/src/util/hooks/useCloudGpxImport.js
Dima-1: Tested. All file types are imported correctly

map/src/util/hooks/useGpxImport.js
Dima-1: equivalent Array.from(fileList || []).filter((file) => isTrackFile(file));

map/src/util/hooks/useLocalGpxImport.js
Dima-1: This is due to refactoring

RZR-UA (approved):

alisa911 (approved):
Remove unnecessary comments

PR #1801, opened 2026-06-17
Add dev-only spatial search toggle in search menu

alisa911:


alisa911:
Screenshot 2026-06-17 at 07 34 37

new
Screenshot 2026-06-17 at 07 46 07

old
Screenshot 2026-06-17 at 07 31 08

PR #1804, opened 2026-06-17
Show matched objects list via info icon in search results

alisa911:


alisa911:
Screenshot 2026-06-17 at 15 27 44

PR #1832, opened 2026-06-26
Speed up search results on large spatial searches

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd/issues/21621

PR #1838, opened 2026-06-29
Implement Copy MVT tile URL in context menu, hotfix crash on disabled WebGL

RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2845

### DIFF
--- a/map/src/map/components/ContextMenu.jsx
+++ b/map/src/map/components/ContextMenu.jsx
@@ -23,6 +23,7 @@ import { ReactComponent as CoordinatesIcon } from '../../assets/icons/ic_action_
 import { ReactComponent as AddPinIcon } from '../../assets/icons/ic_show_on_map_outlined.svg';
 import { ReactComponent as ShowRegionsIcon } from '../../assets/icons/ic_action_world_globe.svg';
 import { ReactComponent as DownloadIcon } from '../../assets/icons/ic_action_gsave_dark.svg';
+import { ReactComponent as CopyIcon } from '../../assets/icons/ic_action_copy.svg';
 import { useTranslation } from 'react-i18next';
 import {
     GLOBAL_GRAPH_HEIGHT_SIZE,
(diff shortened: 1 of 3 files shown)


map/src/map/layers/MvtLayer.js
```diff
@@ -133,6 +137,22 @@ export default function MvtLayer({ config }) {
...
+            window.seIsTilesLoaded = true;
+            console.warn(`${errorLabel}: WebGL is not available`);
+            mtx.setTileURL(osmandTileURL);
+            mtx.setRenderingType(null);
+            ctx.setConfigureMapState((prev) => {
+                const configureMap = {
+                    ...prev,
+                    mapStyle: { tileURL: osmandTileURL, renderingType: null },
+                };
+                updateConfigureMapCache(configureMap);
```
alisa911: The updater passed to setConfigureMapState must be pure, but it writes to cache here and may run twice. Move updateConfigureMapCache(...) out of the updater.

map/src/map/mvt/MvtDemoConfig.js
```diff
@@ -38,3 +38,12 @@ export function isOsmTestTileURL(tileURL) {
 export function isMvtTileURL(tileURL) {
     return tileURL?.type === 'mvt' || isOsmAndTileURL(tileURL) || isOsmTestTileURL(tileURL);
 }
+
+export function isWebGLAvailable() {
```
alisa911, line 44: This creates a WebGL context on every call and never releases it; browsers limit the number of contexts. Cache the result in a module-level variable since WebGL support doesn't change at runtime.

map/src/map/components/ContextMenu.jsx
```diff
@@ -396,6 +402,18 @@ export default function ContextMenu({ setGeocodingData, setRegionData }) {
...
+                                <MenuItem
+                                    id={'se-copy-mvt-tile-url'}
+                                    className={styles.contextMenuItem}
+                                    onClick={() => handleMenuItemClick(copyMvtTileUrl)}
+                                >
+                                    <ListItemIcon className={styles.contextMenuIcon}>
+                                        <CopyIcon />
+                                    </ListItemIcon>
+                                    <ListItemText className={styles.contextMenuItemText}>
+                                        Copy MVT tile URL
```
alisa911, line 414: Hardcoded English string

map/src/map/mvt/MvtDemoConfig.js
RZR-UA: 5ea660aac

map/src/map/components/ContextMenu.jsx
RZR-UA: devel only

map/src/map/layers/MvtLayer.js
RZR-UA: 5ea660aac

PR #1846, opened 2026-06-30
Implement devel-only MvtHybridDemoUnderlay (LLM)

RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2845

### DIFF
--- a/map/src/map/layers/MvtHybridDemo.js
+++ b/map/src/map/layers/MvtHybridDemo.js
@@ -40,10 +40,10 @@ const TemplateTileLayer = L.TileLayer.extend({
 });
 
 export function getHybridUnderlayUrl() {
-    if (sessionStorage.getItem(HYBRID_UNDERLAY_ACTIVE_KEY) !== 'true') {
+    if (localStorage.getItem(HYBRID_UNDERLAY_ACTIVE_KEY) !== 'true') {
         return '';
     }
-    return sessionStorage.getItem(HYBRID_UNDERLAY_URL_KEY) || '';
+    return localStorage.getItem(HYBRID_UNDERLAY_URL_KEY) || '';
 }
 
 function emitHybridUnderlayUrlChange() {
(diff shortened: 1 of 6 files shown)


map/src/map/layers/MvtHybridDemo.js
```diff
@@ -0,0 +1,149 @@
...
+    const mtx = useContext(MapContext);
+    const underlayUrl = useHybridUnderlayUrl();
+
+    useEffect(() => {
+        if (!underlayUrl || !isOsmAndTileURL(mtx.tileURL)) {
+            return undefined;
+        }
+
+        const pane = map.getPane(HYBRID_UNDERLAY_PANE) || map.createPane(HYBRID_UNDERLAY_PANE);
+        pane.style.zIndex = '200';
```
alisa911: z-index 200 collides with Leaflet's default tilePane (also 200), so stacking order vs the base tiles is DOM-order dependent. Use a value strictly between base (200) and mvtDemoPane (250), e.g. 230

map/src/map/layers/MvtHybridDemo.js
```diff
@@ -0,0 +1,149 @@
...
+}
+
+function deactivateHybridUnderlayUrl() {
+    sessionStorage.removeItem(HYBRID_UNDERLAY_ACTIVE_KEY);
+    emitHybridUnderlayUrlChange();
+}
+
+function subscribeHybridUnderlayUrl(callback) {
+    window.addEventListener(HYBRID_UNDERLAY_EVENT, callback);
+    window.addEventListener('storage', callback);
```
alisa911: Dead listener: storage events don't fire for sessionStorage in the same tab, and sessionStorage isn't shared across tabs. Only the custom event does the work — can be removed.

map/src/map/layers/MvtHybridDemo.js
```diff
@@ -0,0 +1,149 @@
...
+
+    return () => {
+        window.removeEventListener(HYBRID_UNDERLAY_EVENT, callback);
+        window.removeEventListener('storage', callback);
+    };
+}
+
+export function useHybridUnderlayUrl() {
+    return useSyncExternalStore(subscribeHybridUnderlayUrl, getHybridUnderlayUrl, () => '');
```
alisa911, line 96: getHybridUnderlayUrl as getSnapshot hits sessionStorage on every render of every subscriber. Minor — could cache.

map/src/map/layers/MvtHybridDemo.js
```diff
@@ -0,0 +1,149 @@
...
+    const map = useMap();
+    const mtx = useContext(MapContext);
+    const underlayUrl = useHybridUnderlayUrl();
+
+    useEffect(() => {
+        if (!underlayUrl || !isOsmAndTileURL(mtx.tileURL)) {
+            return undefined;
+        }
+
+        const pane = map.getPane(HYBRID_UNDERLAY_PANE) || map.createPane(HYBRID_UNDERLAY_PANE);
```
alisa911: Duplicates the generic pane setup added to MvtLayer (config.pane/paneZIndex). Consider sharing a helper instead of repeating createPane + zIndex + pointerEvents.

RZR-UA: b99b66f71

map/src/map/layers/MvtHybridDemo.js
RZR-UA: b99b66f71

map/src/map/layers/MvtHybridDemo.js
RZR-UA: b99b66f71

map/src/map/layers/MvtHybridDemo.js
RZR-UA: b99b66f71

alisa911 (approved):

PR #1861, opened 2026-07-08
Show selected search pin when result is off-screen

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/3265

PR #1863, opened 2026-07-08
Decode OsmGpx track geometry on the client and sort routes by date

alisa911:


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2708

RZR-UA (approved):

PR #1876, opened 2026-07-15
Actualize tests 23,24,35 (back-to-search-results)

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/25331

### DIFF
--- a/tests/selenium/src/actions/map/actionCheckPoi.mjs
+++ b/tests/selenium/src/actions/map/actionCheckPoi.mjs
@@ -7,18 +7,21 @@ async function waitByPoiMarker({ iconWpt, colorIcon, name }) {
     const marker = By.css(`[se-poi-name="${name}"]`);
     return await enclose(
         async () => {
-            const icons = await driver.findElements(icon);
-            if (icons?.length > 0) {
-                return icons[0];
-            }
-            const markers = await driver.findElements(marker);
-            if (markers?.length > 0) {
-                return markers[0];
+            const elements = [...(await driver.findElements(icon)), ...(await driver.findElements(marker))];
+            for (const element of elements) {
+                if ((await element.getCssValue('visibility')) !== 'hidden') {
+                    return element;
+                }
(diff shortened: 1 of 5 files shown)


tests/selenium/src/actions/map/actionCheckPoi.mjs
```diff
@@ -1,5 +1,25 @@
-import { waitBy, waitByRemoved } from '../../lib.mjs';
+import { enclose, waitBy, waitByRemoved } from '../../lib.mjs';
 import { By } from 'selenium-webdriver';
+import { driver } from '../../options.mjs';
+
+async function waitByPoiMarker({ iconWpt, colorIcon, name }) {
+    const icon = By.id(`se-poi-marker-icon-${iconWpt}-${colorIcon}-${name}`);
+    const marker = By.css(`[se-poi-name="${name}"]`);
```
alisa911, line 7: waitByPoiMarker doesn't skip visibility:hidden elements the way waitBy does

tests/selenium/src/actions/map/actionCheckPoi.mjs
alisa911, line 37: The positive check now accepts icon or [se-poi-name], but this removal check still only waits for the icon id. Asymmetric — a lingering [se-poi-name="…"] marker would pass the hidden assertion. Consider mirroring the 2-way logic here.

tests/selenium/src/actions/map/actionCheckPoi.mjs
RZR-UA: 50e482207e7d30fd53250f711c3665ec28f6855b

tests/selenium/src/actions/map/actionCheckPoi.mjs
RZR-UA: 50e482207e7d30fd53250f711c3665ec28f6855b

alisa911 (approved):

PR #1879, opened 2026-07-16
Tag filter

Dima-1:
https://github.com/osmandapp/OsmAnd/issues/24809

### DIFF
--- a/map/src/infoblock/components/wpt/WptTagsProvider.js
+++ b/map/src/infoblock/components/wpt/WptTagsProvider.js
@@ -221,13 +221,19 @@ async function getWptTags(obj, type, ctx) {
 
         const poiNameTags = {};
         const mainTags = {};
-        Object.entries(tags).forEach(([key, value]) => {
+        for (const [key, value] of Object.entries(tags)) {
             if (key.startsWith(POI_NAME) && value) {
                 poiNameTags[key] = value;
+            } else if (key.includes(WIKIDATA) || key.includes(WIKIPEDIA)) {
+                const tagKey = key.includes(WIKIDATA) ? WIKIDATA : WIKIPEDIA;
+                const tagObj = await buildTagObj({ key: tagKey, value: value, lang: null, ctx, subtypeTag });
+                if (tagObj) {
+                    res.push(tagObj);
+                }
             } else {
                 mainTags[key] = value;
             }
-        });
+        }
 
         const poiNameTagObj = await buildPoiNameTagObj(poiNameTags, ctx, subtypeTag);
         if (poiNameTagObj) {
--- a/map/src/infoblock/components/wpt/WptTagsProvider.js
+++ b/map/src/infoblock/components/wpt/WptTagsProvider.js
@@ -52,6 +52,7 @@ const EMAIL = 'email';
 const WEBSITE = 'website';
 const CUISINE = 'cuisine';
 export const CUISINE_PREFIX = 'cuisine_';
+const ROUTE = 'route';
 export const IMAGE_OSM_TAG = 'image';
 export const MAPILLARY_OSM_TAG = 'mapillary';
 export const WIKIDATA = 'wikidata';
@@ -64,7 +65,6 @@ export const ICON_NAME_EXTENSION = 'icon';
 export const BACKGROUND_TYPE_EXTENSION = 'background';
 export const PROFILE_TYPE_EXTENSION = 'profile';
 export const ADDRESS_EXTENSION = 'address';
-export const AMENITY_ORIGIN_EXTENSION = 'amenity_origin';
 export const NAME = 'name';
 export const EN_NAME = 'en_name';
 export const ALT_NAME = 'osm_tag_alt_name';
@@ -108,14 +108,9 @@ const HIDDEN_EXTENSIONS = [
     BACKGROUND_TYPE_EXTENSION,
     PROFILE_TYPE_EXTENSION,
     ADDRESS_EXTENSION,
-    AMENITY_ORIGIN_EXTENSION,
-    AMENITY_PREFIX + NAME,
-    AMENITY_PREFIX + TYPE,
-    AMENITY_PREFIX + SUBTYPE,
 ];
 
 const HIDDEN_EXTENSIONS_POI = [
-    ...HIDDEN_EXTENSIONS,
     ICON_KEY_NAME,
     POI_ICON_NAME,
     TYPE_OSM_TAG,
@@ -231,7 +226,7 @@ async function getWptTags(obj, type, ctx) {
             }
         }
 
-        tags = fixTagsKeys(tags);
+        tags = filterWebKeys(tags);
         tags = await filterTagsByVisibility(tags);
         for (const [key, value] of Object.entries(tags)) {
             if (!shouldSkipKey(key)) {
@@ -400,38 +395,40 @@ export async function addPoiTypeTag({
 async function filterTagsByVisibility(tags) {
     if (Object.keys(tags).length === 0) return tags;
 
-    try {
-        let response = await apiPost(`${process.env.REACT_APP_USER_API_SITE}/search/get-tags-visibility`, tags, {
-            apiCache: true,
-        });
+    let response = await apiPost(`${process.env.REACT_APP_USER_API_SITE}/search/get-tags-visibility`, tags, {
+        apiCache: true,
+    });
 
-        if (!response?.data || typeof response.data !== 'object' || Array.isArray(response.data)) {
-            return tags;
-        }
-
+    if (response?.data && typeof response.data === 'object' && !Array.isArray(response.data)) {
         return response.data;
-    } catch {
-        return tags;
     }
+
+    return fixTagsKeysFallback(tags);
 }
 
-function fixTagsKeys(tags) {
+function fixTagsKeysFallback(tags) {
     let res = {};
     for (const [key, value] of Object.entries(tags)) {
         let newKey = key;
         if (key === AMENITY_PREFIX + OPENING_HOURS) {
             newKey = key.replace(AMENITY_PREFIX, '');
-        } else if (
-            key.startsWith(AMENITY_PREFIX) ||
-            key.startsWith(ALT_NAME) ||
-            HIDDEN_EXTENSIONS.includes(key) ||
-            HIDDEN_EXTENSIONS_POI.includes(key)
-        ) {
+        } else if (key.startsWith(AMENITY_PREFIX) || HIDDEN_EXTENSIONS.includes(key)) {
             continue;
         } else {
             newKey = key.replace(OSM_PREFIX, '');
         }
         res[newKey] = value;
+    }
+    return res;
+}
+
+function filterWebKeys(tags) {
+    let res = {};
+    for (const [key, value] of Object.entries(tags)) {
+        if (key.startsWith(ALT_NAME) || HIDDEN_EXTENSIONS_POI.includes(key)) {
+            continue;
+        }
+        res[key] = value;
     }
     return res;
 }
@@ -632,7 +629,14 @@ function getWikipediaURL(key, value) {
 }
 
 function shouldSkipKey(key) {
-    return key === 'idObj' || key === 'note';
+    return (
+        key === 'idObj' ||
+        key === 'name' ||
+        key === 'subway_region' ||
+        key === 'note' ||
+        key === 'lang_yes' ||
+        key.includes(ROUTE)
+    );
 }
 
 export function openWikipediaContent(tag, setDevWikiContent) {
--- a/map/src/infoblock/components/wpt/WptTagsProvider.js
+++ b/map/src/infoblock/components/wpt/WptTagsProvider.js
@@ -60,12 +60,14 @@ const INSTAGRAM = 'instagram';
 export const OSM_PREFIX = 'osm_tag_';
 const COLLAPSABLE_PREFIX = 'collapsable_';
 export const COLOR_NAME_EXTENSION = 'color';
+const ICON_NAME_EXTENSION = 'icon';
 export const BACKGROUND_TYPE_EXTENSION = 'background';
 export const NAME = 'name';
 export const EN_NAME = 'en_name';
 export const ALT_NAME = 'osm_tag_alt_name';
 const ORIGINAL_ICON = 'originalIcon';
-const SVG = 'svg';
+const MARKER_ICON_SVG = 'svg';
+const MARKER_OBJ_ID = 'idObj';
 
 // from Amenity object
 export const POI_NAME = WEB_POI_PREFIX + 'name';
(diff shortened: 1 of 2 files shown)


map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -109,10 +110,6 @@ const HIDDEN_EXTENSIONS = [
     BACKGROUND_TYPE_EXTENSION,
     PROFILE_TYPE_EXTENSION,
     ADDRESS_EXTENSION,
-    AMENITY_ORIGIN_EXTENSION,
-    AMENITY_PREFIX + NAME,
-    AMENITY_PREFIX + TYPE,
-    AMENITY_PREFIX + SUBTYPE,
 ];
 
 const HIDDEN_EXTENSIONS_POI = [
```
alisa911: Rename HIDDEN_EXTENSIONS_POI to web tags, and HIDDEN_EXTENSIONS should not exist, it duplicates the server logic

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -232,145 +229,211 @@ async function getWptTags(obj, type, ctx) {
...
-                if (tagObj.key.startsWith(POI_NAME)) {
-                    tagObj.key = tagObj.key.replace(POI_NAME, 'shared_string_name');
-                    tagObj.textPrefix = tagObj.key;
-                }
-
-                if (key.includes(WIKIDATA)) {
-                    tagObj = addWikidataTags(key, value, tagObj);
-                }
-                res.push(tagObj);
+        const poiNameTags = {};
```
alisa911, line 220: Unclear name remainingTags remaining after what, filterTagsByVisibility does not filter, the fallback with duplication is not needed, pick one thing either mainTag or mainEntry, const SVG is a magic constant

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -232,145 +229,211 @@ async function getWptTags(obj, type, ctx) {
...
     return { res, id, type: typeTag, subtype: subtypeTag };
 }
 
-function mergeTagsWithLang(tags) {
-    tags.forEach((tag) => {
-        if (tag.key.includes(':')) {
-            const arr = tag.key.split(':');
-            tag.key = arr[0];
-            tag.lang = arr[1];
+async function buildTagObj(key, value, lang, ctx, subtypeTag) {
```
alisa911: buildTagObj has a large number of arguments, use {}

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -232,145 +229,211 @@ async function getWptTags(obj, type, ctx) {
...
-                    case CUISINE:
-                        tagObj.icon = <CuisineIcon />;
-                        break;
-                    case SERVICE_TIMES:
-                    case COLLECTION_TIMES:
-                        tagObj.icon = <TimeIcon />;
-                        tagObj.needLinks = false;
-                        tagObj.textPrefix = value;
-                        break;
-                    case OPENING_HOURS:
```
alisa911, line 266: If you touched the switch in buildTagObj, then rewrite it properly

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -232,145 +229,211 @@ async function getWptTags(obj, type, ctx) {
...
+                hasCuisine = true;
+            }
+            res.push(tagObj);
         }
         if (hasCuisine) {
             res = res.filter((t) => t.key !== CUISINE);
         }
     }
 
-    res = res.filter((t) => !t.key.startsWith(WEB_PREFIX));
```
alisa911, line 350: Why did you remove the WEB_PREFIX filter?

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -397,18 +460,32 @@ export async function addPoiTypeTag({
     return tagObj;
 }
 
-function fixTagsKeys(tags) {
+async function filterTagsByVisibility(tags) {
```
alisa911: The server returns an untyped map where a value is either a string or a nested localizations object. With no schema, the client has to rebuild what the server already computed: groupLocalizedTags re-splits the keys, re-picks the main language and rebuilds otherLangs. That double interpretation is where the bugs came from, and if the server changes the shape the web breaks silently.

Fix: return a typed, render ready list like { key, value, lang, otherLangs } already grouped, so the client only renders it.

alisa911 (commented):
The goal was "one filtering logic". In practice only the predicates (shouldDisplayKey, isKeyToSkip, getPoiAdditionalType) were moved into the shared AdditionalInfoBundle, while the loop that assembles "which tags to show" was duplicated: one copy in SearchService.checkTagsVisibility, a second in Android's AmenityUIHelper.buildInternal, and a third (partial) one in the web fallback fixTagsKeysFallback plus the extra work in groupLocalizedTags. Three sources of truth instead of one.

Move the whole logic into a method in AdditionalInfoBundle. The web should receive a ready to display list of tags, and there should be no groupLocalizedTags or other such logic on the web. The web should only build the UI, which icon, a list or a tag, and so on.

alisa911:
This violates map/CLAUDE.md

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/007035872f479307cd584a4ecfd4da5b822d82cb

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/64adefa51cc79e6436aad814d50d6ba915766307

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/64adefa51cc79e6436aad814d50d6ba915766307

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/2186facf4890fa990b757be924547b2afe62bccc

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/2186facf4890fa990b757be924547b2afe62bccc

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/2186facf4890fa990b757be924547b2afe62bccc

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -418,6 +495,23 @@ function fixTagsKeys(tags) {
...
 
+function filterWebKeys(tags) {
+    let res = {};
+    for (const [key, value] of Object.entries(tags)) {
+        if (
+            key.startsWith(ALT_NAME) ||
+            HIDDEN_EXTENSIONS_POI.includes(key) ||
+            key.startsWith(ORIGINAL_ICON) ||
+            key === SVG
+        ) {
```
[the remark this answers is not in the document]
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/2186facf4890fa990b757be924547b2afe62bccc

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -397,24 +427,37 @@ export async function addPoiTypeTag({
     return tagObj;
 }
 
-function fixTagsKeys(tags) {
+async function fetchVisibleTags(tags) {
```
alisa911, line 418: If the request fails, this returns [] and the info block silently loses all tags, fall back to showing raw tags.

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -232,145 +219,188 @@ async function getWptTags(obj, type, ctx) {
...
+            if (entry.otherLangs?.length) {
+                tagObj.otherLangs = await buildOtherLangTags(entry.otherLangs, ctx, subtypeTag);
             }
+
+            if (tagObj.collapsable && tagObj.textPrefix === CUISINE) {
+                hasCuisine = true;
+            }
+            res.push(tagObj);
         }
         if (hasCuisine) {
```
alisa911: This hasCuisine filter can never trigger — the backend already drops the plain cuisine tag when a cuisine/dish group exists (containsAny in getVisibleTags), so the web never receives both; remove it or keep it with a comment explaining it's defensive.

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -232,145 +219,188 @@ async function getWptTags(obj, type, ctx) {
...
+}
+
+async function buildPoiNameTagObj(poiNameTags, ctx, subtypeTag) {
+    const entries = Object.entries(poiNameTags).map(([key, value]) => {
+        const [base, lang] = key.split(':');
+        return lang ? { key: base, value, lang } : { key: base, value };
     });
-    return tags;
+
+    const mainEntry = entries.find((entry) => entry.lang) ?? entries[0];
```
alisa911: The main name is just the first entry with a lang, not the user's language — and this duplicates the same arbitrary policy the server uses in groupLocalizedTags; the language choice should live in one place and use the user's locale.

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -232,145 +219,188 @@ async function getWptTags(obj, type, ctx) {
...
+    if (!mainEntry) {
+        return null;
+    }
+
+    const tagObj = await buildTagObj({
```
alisa911, line 373: buildOtherLangTags awaits buildTagObj one by one — Promise.all would build them in parallel.

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -397,24 +427,37 @@ export async function addPoiTypeTag({
...
-function fixTagsKeys(tags) {
+async function fetchVisibleTags(tags) {
+    if (Object.keys(tags).length === 0) return [];
+    tags = Object.fromEntries(
+        Object.entries(tags).map(([key, value]) => [
+            key,
+            typeof value === 'number' || typeof value === 'boolean' ? String(value) : value,
+        ])
+    );
+    let response = await apiPost(`${process.env.REACT_APP_USER_API_SITE}/search/filter-visible-tags`, tags, {
```
alisa911: response is never reassigned — const instead of let.

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -232,145 +219,188 @@ async function getWptTags(obj, type, ctx) {
...
+            } else if (key.includes(WIKIDATA) || key.includes(WIKIPEDIA)) {
+                const tagKey = key.includes(WIKIDATA) ? WIKIDATA : WIKIPEDIA;
+                const tagObj = await buildTagObj({ key: tagKey, value: value, lang: null, ctx, subtypeTag });
+                if (tagObj) {
+                    res.push(tagObj);
                 }
+            } else {
+                mainTags[key] = value;
+            }
+        }
```
alisa911, line 235: stray double blank line left after the refactor.

alisa911 (commented):
Summary:

To make the tag order identical on web and Android, two things are needed:

1. Add a shared sort function to AmenityRowsBuilder that orders rows by
(order, key). It must be the single sorting point: the server and Android
both call it, and nothing reorders rows after it — today Android breaks
ties by name, which is only filled on Android, so the server sorts
differently.

2. Use LinkedHashMap instead of HashMap everywhere along the path
(e.g. getFilteredInfo) — otherwise the same algorithm returns a different
order on every run.

Note: for fully identical results the endpoint also needs a lang
parameter (the main localized value and description order depend on the
user language), and the web should render rows strictly in the response
order.

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/807cd62bb807204ccf15894ec14716df20b2cab0

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/807cd62bb807204ccf15894ec14716df20b2cab0

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/807cd62bb807204ccf15894ec14716df20b2cab0

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/807cd62bb807204ccf15894ec14716df20b2cab0

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/807cd62bb807204ccf15894ec14716df20b2cab0

map/src/infoblock/components/wpt/WptTagsProvider.js
Dima-1: https://github.com/osmandapp/web/pull/1879/commits/af76415b07e087941790d18f406ed8edade136e1

map/src/infoblock/components/wpt/WptTagsProvider.js
```diff
@@ -397,24 +415,38 @@ export async function addPoiTypeTag({
...
+    tags = Object.fromEntries(
+        Object.entries(tags).map(([key, value]) => [
+            key,
+            typeof value === 'number' || typeof value === 'boolean' ? String(value) : value,
+        ])
+    );
+    const response = await apiPost(`${process.env.REACT_APP_USER_API_SITE}/search/visible-tags`, tags, {
+        apiCache: true,
+        params: { lang: i18n.language },
+    });
```
RZR-UA, line 429: It's better to pre-filter visible tags in the results of `search/search-poi?locale=uk&lat=50.4043&lon=30.5254&baseSearch=false&timeZone=Europe%2FBrussels` instead of flooding the server with this `visible-tags` API on every Amenity card display.

RZR-UA (approved):

RZR-UA (approved):

Issue #1893, opened 2026-07-22
Web - Spatial Search & Format Results


vshcherb:
# To be designed (Follow up task)
- [ ] Cells formatting

# Bugs

- [x] Fix broken category translations (bank = ok, банк = failed)
- [x] Change `type` in `map/poi/?name=X&type=Gas+station` from EN to XML key
- [x] Check why top_index_brand_сільпо takes a lot of time to show results on zoom 4-5
- [x] Increase POI_CATEGORY_PREFIX_LENGTH for `#^top_index_xxxxx_xxx` (19? 21? chars)
- [x] Broken search POI categories if translated name is longer than English one (Дитячий садок) [commit](https://github.com/osmandapp/OsmAnd/commit/0b7e2da8750025be6157c8d95c8c4fbcb227bc08#diff-47f383831d7dfa0f75e5f77107dfb3378c653afc2fa1e3504e63f3108c3711cdR108)
- [x] `Bratislava billa` is displayed sometimes as `Bratislava strada` (when searched from Kyiv)

# Web tasks
- [x] Highlight ref matching, interpolation (somehow) with braces
- [x] Intersection of 2 Amenities or Amenity is not first object
- [x] Poi category + geo object (name of geobject, dist? and relocate)
- [x] Poi translation provider
- [x] Multithread pool + Cancel - Production - fine tune
- [x] Use new search by poi category (much faster for supported) categories
- [x] Autosuggestions (try and test first?) web branch `rzr-autocomplete`

From https://github.com/osmandapp/OsmAnd/issues/25331

dmpr0:
## Review

<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th></th>
      </tr>
      <tr>
        <td><img src="https://github.com/user-attachments/assets/9c9730ef-e003-4d3f-a8ab-14da9c039164"  width="600" height="auto"/> </td>
      </tr>
    </table>
</details>  

- [ ] Missing POI category suggestions, for example try to search for `Дитячий садок` or `Hardware store`

- - - - - -

<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th></th>
      </tr>
      <tr>
        <td><video src="https://github.com/user-attachments/assets/4f88c614-9939-4798-a5f9-e4f3d203b026"  width="300" height="auto"/> </td>
      </tr>
    </table>
</details>  

- [x] Clicking on result on the map open wrong context menu, for coordinates not for POI

[...]

PR #1909, opened 2026-07-29
Drop client zoom gate for POI categories, render server zoom-in hint

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1893

PR #1912, opened 2026-07-29
Fix typing suggestions not shown on main search page

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1893

PR #1932, opened 2026-08-12
Change text

TinaPohor:
https://github.com/osmandapp/OsmAnd-iOS/issues/5274


main/docs/user/map/tracks/track-context-menu.md
EugeneZmeuk: https://osmand.net/docs/user/map/tracks/track-context-menu?current-os=android#points--waypoints

We need to change screenshots of the Android and iOS context menu with these waypoints:  showing the map with a track and waypoints. While its show only the Context menu.

Issue #1849, opened 2026-07-01
Missing author data for Image in POI photos


dmpr0:
## Description
When viewing the details of an artwork POI, the relevant information such as the author's name and the object description is not displayed.

### How to reproduce
* Navigate to https://osmand.net/map/poi/?name=Na%C5%A1+tovor&type=Artwork&pin=43.161537,16.694315#20/43.16150/16.69413

### Actual result
<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/411872a9-a807-49a2-b0e0-134d70ecf624"  width="300" height="auto"/> </td>
  </tr>
</table>   

* There is no information regarding the **Author** or **Description**.

### Expected result
* Show all available information for the image, including the **Author** and **Description**.

alisa911:
<img width="1387" height="841" alt="Image" src="https://github.com/user-attachments/assets/f7f14fab-b290-44d0-8c32-7bee33a54cde" />

Zirochkabila:
# Testing 1

> Web - prod

| # | Test Scenario | POI / URL | Expected Result | Actual Result | Status |
| :--- | :--- | :--- | :--- | :--- | :---: |
| **TP-01** | Artwork Author Display | `Naš tovor` (Artwork POI) | Author/Artist field is present and visible in POI details | Author field is correctly displayed | 🟢 **PASSED** |
| **TP-02** | Artwork Description Display | `Naš tovor` (Artwork POI) | Description field is present and displays object details | Description is visible in details card | 🟢 **PASSED** |
| **TP-03** | Human-readable Field Names | `Artwork` POI type | Attribute titles are clean without raw OSM keys | Attribute labels are clean and human-readable | 🟢 **PASSED** |

dmpr0:
<table style="width:100%">
  <tr>
    <th>Android</th>
    <th>iOS</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/d89ba0ab-3017-448d-bd13-e9cbdcedf124"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/59c89077-95b4-4947-b424-cddf3b5f274d"  width="300" height="auto"/> </td>
  </tr>
</table>   


Android - Ok
iOS – Not ok. Does it require the iOS Master branch?

PR #1948, opened 2026-08-24
Fix photos lost after tag filter refactoring

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1849

RZR-UA (approved):

Issue #1957, opened 2026-08-27
Shared favorite folder issues


dmpr0:
<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th></th>
      </tr>
      <tr>
        <td><video src="https://github.com/user-attachments/assets/faa5a50d-bcb7-4883-9886-55b7f18e9f09"  width="300" height="auto"/> </td>
      </tr>
    </table>
</details>  

- [x] Unable to open context menu for the point from the Shared folder
- [x] Map behaviour should be the same as for Favorite folders. In this case map stays over Kyiv, but all points are located in Paris

- - - - - -
 
<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th></th>
      </tr>
      <tr>
        <td><video src="https://github.com/user-attachments/assets/4d72d0ed-f55e-497d-a2f3-165adb10c74f"  width="300" height="auto"/> </td>
      </tr>
    </table>
</details>  

- [x] Crash after trying to access shared folder from My Places

- - - - - -
 
- [x] Public folder should be visible without login

Zirochkabila:
# Testing 1

> Web - prod

| # | Test Scenario | Area / Action | Expected Result | Actual Result | Status |
| :--- | :--- | :--- | :--- | :--- | :---: |
| **TP-01** | Public Folder Access | Logged out / Incognito | Folder & points visible via direct link without login | Accessible without login | 🟢 **PASSED** |
| **TP-02** | My Places Shared Folder Access | My Places -> Shared Folders | Folder opens smoothly without crash | Opens correctly without crashes | 🟢 **PASSED** |
| **TP-03** | Map Recentering / Fit Bounds | Open folder distant from current view | Map automatically centers and scales to folder points | Auto-fits bounds to folder points | 🟢 **PASSED** |
| **TP-04** | Shared Point Context Menu | Click on a shared point | Clicking on a shared point successfully opens its context card | Context menu opens properly | 🟢 **PASSED** |

PR #1958, opened 2026-08-27
Fix share link

alisa911:
…alone in menu


alisa911:
https://github.com/osmandapp/web/issues/1957

PR #1975, opened 2026-09-02
Map: draw HH alternative routes

vshcherb:
Draws the alternative routes that HH routing returns. Needs the `alternatives` parameter of `/routing/route` from osmandapp/OsmAnd-tools#hh-alt-routes-plateau, which in turn needs osmandapp/OsmAnd#25805. Ref osmandapp/OsmAnd-Issues#2843.

## What it does

- `calculateRoute` asks for `ROUTE_ALTERNATIVES` (2) alternatives and gives the features that come back with `properties.alternative` their own style: violet at 0.5 opacity, weight 8. The colour is deliberately one the OSM base map does not use for roads — grey blends in and the lines stop reading as routes.
- **Hover** shows what picking an alternative would cost, compared against the route currently on the left:

  `Route 2. 27.5 km (2.0 km longer), 35 min (1 min shorter), 2044 cost (+163)`

  The cost is only shown with `REACT_APP_DEVEL_FEATURES`.
- **Click** puts that route on the left: it moves to the front of the collection — the summary card, the elevation graph and the saved track all read the first feature — and the two exchange the `alternative` marker and their styles. It is matched by that marker rather than by object identity, since the layer may hold a copy of the feature.
- Alternatives are display only, so `convertRouteToTrack()` skips whatever still carries the marker. Without that their geometry gets appended to the track being edited and to the elevation graph, since it takes every LineString of the collection.

## Testing a router change locally

`USE_LOCAL_ROUTING=yes` (`yarn start:local-routing`) proxies `/routing/` to `localhost:8080` and leaves tiles, search and the rest on test.osmand.net. `start:local` sends everything local, which means no tiles while a routing change is being looked at.

Verified through the dev proxy against a local server over `Ukraine_kyiv_europe_2.obf`, Podil -> Teremky: two alternatives, 16.2 km (+0.5%) and 27.5 km (+9.4%).

---

## AI disclaimer

Written with Claude Code (Claude Opus 5), driven by @vshcherb. Summary of the requests it worked from, in order:

1. Analyse how to do alternative routes in OsmAnd; `HHRoutePlanner` is the main class; read the fast-routing blog post to see how HH works. What is wanted is *interesting, correct* alternatives — maximally different geometry at a similar time, not a detour around a pole for +5%. 1–2 alternatives always, and it has to be fast. The existing ideas are tied to opaque heuristics in the code; wanted an understandable one, roughly Google-like but not a black box, so it can be tuned on user feedback later. Run tests on the local maps, and decide what to compare against and which criteria to use.
2. Turn it into code, and draw the alternatives so they can be tested with HH.
3. Not the Android app — the thing to run is OsmExtractionUI (OsmAndMapCreator, tools project), `MapRouterLayer`.
4. Both routes have the same colour — give the second one a different colour. Which one is the optimal?
5. Too many labels on alt 1 / alt 2, they flood the map. Use green for alt 1 and magenta for alt 2. Reverse the draw order — alt 2 first, then alt 1, then the main route, so the main route is always fully visible on top.
6. Open pull requests from the tools and android branches.

All numbers in this description are measurements from those runs, not estimates.


map/src/map/layers/NavigationLayer.js
```diff
@@ -18,9 +18,12 @@ import {
 import { NAVIGATE_URL } from '../../manager/GlobalManager';
 import { navigationObject } from '../../store/navigationObject/navigationObject';
 import { pickNextRoutePoint } from '../../manager/NavigationManager';
-import { POINT_MARKER_Z_INDEX_OFFSET, TURN_DOT_Z_INDEX_OFFSET } from '../util/ZIndexes';
+import { ALTERNATIVE_ROUTE_STYLE } from '../../store/geoRouter/legacy/calculateRoute';
 
 const DRAG_DEBOUNCE_MS = 10;
+const ALTERNATIVE_HOVER_OPACITY = 0.9;
+const DEVEL_FEATURES = process.env.REACT_APP_DEVEL_FEATURES === 'yes';
```
alisa911: ctx.develFeatures already exists in AppContext and is toggleable at runtime, use it

map/src/map/layers/NavigationLayer.js
```diff
@@ -353,7 +356,71 @@ const NavigationLayer = ({ geocodingData, region }) => {
...
+            const d = Math.round(alt.routingTime - (main.routingTime ?? 0));
+            const delta = main.routingTime ? ` (${d > 0 ? '+' : ''}${d})` : '';
+            parts.push(`${Math.round(alt.routingTime)} cost${delta}`);
+        }
+        return `Route ${props.alternative}. ` + parts.join(', ');
+    };
+
+    // Show the picked alternative in place of the route on the left: it moves to the front of the
+    // collection (the summary is read from the first feature) and the two swap markers and styles.
+    const selectAlternative = (feature) => {
```
alisa911, line 404: After swapping, turn/description point features still belong to the previous main route

alisa911, line 404: The header text is not refreshed after a swap

map/src/map/layers/NavigationLayer.js
```diff
@@ -353,7 +356,71 @@ const NavigationLayer = ({ geocodingData, region }) => {
         fillOpacity: 0.8,
     };
 
+    // "Route 2. 27.5 km (5.1 km shorter), 35 min (7 min longer), 1800 cost (+500)"
+    // Everything is compared against the route currently shown on the left.
+    const describeAlternative = (props) => {
```
alisa911, line 367: Tooltip strings and units are hardcoded in English

map/src/map/layers/NavigationLayer.js
```diff
@@ -353,7 +356,71 @@ const NavigationLayer = ({ geocodingData, region }) => {
...
+        routeObject.putRoute({ route: { ...route, features } });
+    };
+
     const onEachFeature = ({ feature, layer, id = null }) => {
+        if (feature.properties?.alternative) {
+            // translucent so the main route stays readable, but thick enough to click
+            layer.bindTooltip(describeAlternative(feature.properties), { sticky: true });
+            layer.on('mouseover', () => layer.setStyle({ opacity: ALTERNATIVE_HOVER_OPACITY }));
+            layer.on('mouseout', () => layer.setStyle({ opacity: ALTERNATIVE_ROUTE_STYLE.opacity }));
+            layer.on('add', () => layer.bringToBack());
```
alisa911: bringToBack() pushes the alternative below all overlays including visible tracks

map/src/map/layers/NavigationLayer.js
alisa911, line 549: fitBounds now includes alternative geometry

alisa911:
## AI disclaimer

Reviewed and fixed with Claude (Claude Fable 5), driven by @alisa911. What it was asked to do, in order — corrections included:

- Read issue #2855 and go through the web and tools PRs, plus the rest for context. No guessing.
- Its first pass came back as a summary instead of a review, so it got sent back to do a real one. The review comments here and in OsmAnd-tools#1664 are the result.
- It didn't notice the PR was adding an env constant when `ctx.develFeatures` already existed — that had to be pointed out. Pushed again on NavigationLayer and two more issues surfaced: clicks on an alternative bubble up to the map click handler, and on touch there's no hover, so a tap swaps the route with no warning.
- One of its own review comments, about the header text, turned out to describe state that nothing renders anymore. So the task became: delete that dead code.
- Then the review findings got fixed one by one: switch to `ctx.develFeatures`, hide outdated turns after a swap (later replaced by the turns-swap approach), i18n for the tooltip, keep alternatives out of the elevation graph (a Copilot finding it had missed).
- Replacing the hardcoded violet with the profile colour: it produced thirty lines of homemade HSL colour math. Thrown out — drawing the profile colour semi-transparent does the job in three lines.
- Hiding alternatives when the route opens in details or editing took three tries: a filter that misbehaved, then deleting them (so they never came back), then stash-and-restore (broke in the editor, then again coming back from local tracks). What finally worked is simply not rendering them in any track mode.
- When the turns-swap commit landed, the details-hiding quietly stopped working — alternatives had moved into their own layer, out of the filter's reach. It spotted that while reviewing the commit and fixed it again.
- The mystery blue frame around a clicked alternative: the tooltip makes the line focusable, and the browser draws its focus outline around the path's bounding box. Removing tabindex once didn't stick; `outline: none` on the element did.

Everything was reviewed and committed by hand.

PR #1978, opened 2026-09-03
Clear POI overlay markers when all categories are unchecked

### LINKED ISSUE #1945: POI Overlay icons remain visible on the map after the overlay is disabled
## Description
After the **POI Overlay** is disabled, the orange POI icons remain visible on the map until the page is manually refreshed.

### How to reproduce
**Enabling the POI Overlay:**
* Go to https://osmand.net/
* Open **Menu** – **Configure Map** – **POI Overlay**
* Select the checkbox next to **Leisure**
* Click **Apply** — the orange POI icons should appear on the map

**Disabling the POI Overlay:**
* Return to **Configure Map** – **POI Overlay**
* Clear the checkbox next to **Leisure**
* Click **Apply**

### Actual result
<table style="width:100%"
  <tr>
    <td><video src="https://github.com/user-attachments/assets/5361f45d-e968-4b9c-84d9-5e7a4c21dbc2"  width="300" height="auto"/> </td>
  </tr>
</table>

* The orange POI icons remain visible on the map until the page is manually refreshed.

### Expected result
* The icons disappear immediately after the **POI Overlay** is disabled.

vshcherb:
Fixes #1945 — after the POI Overlay is disabled in Configure Map, the orange POI icons stayed on the map until a manual page refresh.

## Cause

`PoiLayer`'s main effect started with:

```js
if ((!isEmpty(ctx.showPoiCategories) && zoom !== prevZoom) || move || isTypeChange) { ... }
else if (isEmpty(ctx.showPoiCategories)) { /* clearPoiList() lives here */ }
```

`move` is the `moveend` timestamp from `useZoomMoveMapHandlers` and is never reset, so it stays truthy after the first map move (which happens on load). The first branch therefore always won; inside it `if (ctx.showPoiCategories.length > 0)` was false, so nothing happened — and the branch that calls `clearPoiList()` when the category list becomes empty was never reached.

## Fix

Check the empty-categories case first, then the zoom/move/type-change refresh. The nested `ctx.showPoiCategories.length > 0` check is now redundant and is dropped.

## Test

New selenium test `search/100-poi-overlay-disable.mjs`: enables the *Cafe and restaurant* category, checks the markers appear, unchecks it, and waits for them to disappear. It fails on `main` (`waitByRemoved` times out on `se-poi-marker-background-#f8931d-circle`) and passes with the fix.

Verified locally against `localhost:3000` (dev server proxied to test.osmand.net): `search` (11), `map` (2) and `menu` (6) suites all pass.

## AI disclaimer

Produced with Claude Code (Opus 5).

Prompts used (summarised):
1. Reproduce and fix issue #1945, using the local web dev server (no postgres) and the local maps folder.
2. Create the pull request, and add a web UI test for this — it is a bad enough problem to be worth covering.

Decided by the agent, not requested explicitly:
- Root-cause choice of fix: reordering the branches in `PoiLayer`'s effect (checking the empty-category case first) rather than resetting the `move` flag after each refresh; the redundant nested `length > 0` check was dropped as part of it.
- Placement, name and shape of the regression test (`tests/selenium/src/tests/search/100-poi-overlay-disable.mjs`), modelled on the existing `91-poi-markers-test.mjs`.
- Running the neighbouring `search`, `map` and `menu` suites as a regression check, and confirming the new test fails on the unfixed code before claiming it covers the bug.


alisa911 (commented):
The test needs to be reworked, it duplicates the existing test and breaks the test order and test conventions. The empty-categories branch in PoiLayer also needs rework

PR #1980, opened 2026-09-04
Cover map/src/util with unit tests

alisa911:
## Cover map/src/util with unit tests

Continues the unit test suite in `tests/unit` — this time the whole `map/src/util` folder.
119 new tests, 135 in total.

### Tests

| File | Covered |
|---|---|
| `01-utils.test.js` | `Utils.js`: distance and bearing, `formatMeters`, `toHHMMSS`, `quickNaNfix`, `sanitizedFileName`, `areSetsEqual`, `createUrlParams`, `encodeString`/`decodeString` round-trip, `truncateText`, `isToday`/`isYesterday` |
| `02-decode-geometry.test.js` | `decodeSimplifiedGeometry.js` |
| `03-date-fmt.test.js` | `dateFmt.js`: `localizeWeekTokens` and every `fmt` formatter |
| `04-app-data-version.test.js` | `appDataVersion.js`: version state and IndexedDB reset paths |
| `05-lang.test.js` | `lang.js` and `LanguageDisplayName.js` |
| `06-gzip.test.js` | `GzipBase64.mjs` and `GzipCompression.js` |

The geometry decoder is checked against an independent encoder that mirrors the server-side
`TrackSimplifyEncoder` (`src/util/fixtures/geometry.js`), so the two sides of the wire format
cannot drift apart unnoticed.

### Fixes found while writing the tests

- `sanitizedFileName` truncated a file name to 255 **bytes** and could cut a multi-byte character
  in half, leaving `<replacement char>` in the name. It now steps back over the UTF-8 continuation bytes,
  and returns early when the name is short enough instead of re-encoding it.
- `createUrlParams` checked `Object.keys(pretty).length` on a string. The result happened to be
  correct; the check now says what it means.

### Notes

- `GzipBase64.mjs` was stubbed in `jest.config.js` (added there only to make the `TracksManager`
  import graph load). The stub is removed — the module pulls in nothing but `pako`.
- `HttpApi.js` and `util/hooks/**` are left uncovered on purpose: the first is the network boundary
  every test already stubs, the second needs `@testing-library/react` and duplicates what the
  selenium suite checks.

### DIFF
--- a/map/src/util/Utils.js
+++ b/map/src/util/Utils.js
@@ -192,9 +192,17 @@ export function cloneTrackObject(track) {
 // Used ONLY when creating a new file
 export function sanitizedFileName(filename, isFavoriteGroup = false) {
     const truncate = (sanitized, length) => {
-        const uint8Array = new TextEncoder().encode(sanitized);
-        const truncated = uint8Array.slice(0, length);
-        return new TextDecoder().decode(truncated);
+        const bytes = new TextEncoder().encode(sanitized);
+        if (bytes.length <= length) {
+            return sanitized;
+        }
+        // step back over the continuation bytes, so a multi-byte character is never cut in half
+        let end = length;
+        while (end > 0 && (bytes[end] & 0xc0) === 0x80) {
+            end--;
+        }
+
+        return new TextDecoder().decode(bytes.slice(0, end));
     };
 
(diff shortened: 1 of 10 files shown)


alisa911:

PR #1983, opened 2026-09-04
Cover track names with unit tests and fix the gpx extension cut

alisa911:


alisa911:

PR #1984, opened 2026-09-04
Cover cloud track with unit tests

alisa911:
## Unit tests for cloud tracks + fixes for the bugs they found

Adds unit-test coverage for the cloud-track managers (names and paths, folder tree, rename, delete, list refresh, `.info` refresh, visible-tracks cache) and fixes every bug found while writing them. Each fix has a test that fails without it.

### Fixed

- **Renaming a nested folder corrupted the path.** `renameFolder` / `renameSmartFolder` used `fullName.replace(folder.name, newName)`, so for `Trips/Trips` or any path where the folder name occurs earlier, the wrong segment was replaced. Both now use the new `renameLastFolderSegment`.
- **`createTrackGroups` returned `undefined` for an empty file list** — `doSort` returns nothing when there is nothing to sort, while every caller expects a list.
- **Deleting a track updated at most one group.** `deleteTracksFromGroups` looked up the group by path and compared the bare file name against `groupFiles` / `files`, which hold the full cloud name, so nothing was spliced and `realSize` was never decremented. Even with the name fixed, the groups above the track (the parent folder, the default group) kept the deleted entry in `files` with a stale `realSize` and last-modified date, and for a root track the default group's `groupFiles` was left untouched whenever that group owns subfolders. Replaced with a recursive walk that removes the track from every group holding it and recomputes `realSize` and the last-modified date there.
- **A file of a nested folder was listed twice.** `addFilesAndCalculateLastModified` pushes the subfolders' files into `group.files` without the duplicate guard used a few lines below, and a folder is visited twice once the default group adopts it as a subfolder. Not visible in the UI (the lists render `groupFiles`), but the tree was wrong.
- **The folder date did not follow a re-saved track.** When a track is saved over an existing one, `updateUpdatetimemsInGroups` bumped the file time but never recalculated `lastModifiedMs` / `lastModifiedDate`, which the folder row displays and date sorting uses.
- **An empty `.info` triggered a pointless merge and redraw** in `applyRefreshedInfoFilesToGpx`; it now behaves like the open-track branch and skips a row without data.
- **A hidden waypoint group came back when the map layer was rebuilt.** `updateGroupsVisibility` updated only `ctx.selectedGpxFile`, while `addTrackToMap` draws waypoints from `ctx.gpxFiles[name].info`; the loaded track is now kept in sync, as the local-track branch already does for `ctx.localTracks`.
- **Visible-tracks cache:** `isVisibleTrack` looked up shared tracks without the `share_visible_marker_` prefix that `updateVisibleCache` writes, so a visible shared track was treated as hidden and dropped from the map when the info block was closed; `updateVisibleCache` threw on an empty cache; `hideAllVisTracks` / `showAllVisTracks` wrote the string `"null"` into localStorage.

### Tests

New and extended suites in `tests/unit/src/tests/tracks/`: `01` (waypoint-group visibility), `04` (folder tree), `05` (rename), `06` (delete), `07` (list refresh), `08` (`.info` refresh), `09` (visible-tracks cache) — 243 tests total, ~9 s.

`jest.config.js`: `useInitialFilesLoad` un-stubbed, `shareConstants` and `VisibleTracks` wired to the real files.

### DIFF
--- a/map/src/manager/SmartFoldersManager.js
+++ b/map/src/manager/SmartFoldersManager.js
@@ -1,5 +1,5 @@
 import { apiGet } from '../util/HttpApi';
-import { filterRegularFolders } from './track/TracksManager';
+import { filterRegularFolders, renameLastFolderSegment } from './track/TracksManager';
 import { updateSortList } from '../menu/actions/SortActions';
 import { SMART_TYPE } from '../menu/share/shareConstants';
 
@@ -80,7 +80,7 @@ export async function deleteSmartFolder(folder, ctx) {
 }
 
 export async function renameSmartFolder(folder, newName, ctx) {
-    const newFolderName = folder.fullName.replace(folder.name, newName);
+    const newFolderName = renameLastFolderSegment(folder.fullName, newName);
     const res = await apiGet(`${process.env.REACT_APP_USER_API_SITE}/mapapi/rename-smart-folder`, {
         params: {
             folderName: folder.fullName,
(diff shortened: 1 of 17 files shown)


alisa911:

PR #1987, opened 2026-09-07
Show cloud request errors instead of swallowing them, fail tests on error dialog

alisa911:
Failed cloud requests (download-file, process-track-data, delete-file, folder/group creation) are now reported to the user via the error dialog instead of being silently swallowed, and the Selenium waitBy fails immediately with that dialog's text instead of hanging until the timeout.


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/3322

RZR-UA (approved):

Issue #1977, opened 2026-09-02
"Save to Cloud" button in route Details doesn't work


DmitryAlexei:
### Description
After building a route on the web map and opening the Details panel, clicking "Save to Cloud" has no effect — the route is not saved and no confirmation dialog/feedback appears.

### Steps to reproduce
Go to osmand.net/map → build a route (set start/destination) → open Details → click Save to Cloud.

### Actual result

https://github.com/user-attachments/assets/46c3fac2-4d04-4408-83c6-28eeccff3359

Expected result: Route should be saved to OsmAnd Cloud

Environment: OsmAnd Web, browser Chrome

Zirochkabila:
# Testing 1

> Web - prod

| # | Test Scenario | Expected Result | Actual Result | Status |
| :--- | :--- | :--- | :--- | :---: |
| **TP-01** | Save Route to Cloud | Clicking "Save to Cloud" in Route Details triggers save process and feedback modal | Route successfully saved to Cloud with feedback/confirmation | 🟢 **PASSED** |
| **TP-02** | Cloud Availability after Save | Saved route GPX appears under OsmAnd Cloud -> Tracks on Web and Mobile apps | Route appears in Cloud tracks and syncs seamlessly to mobile apps | 🟢 **PASSED** |

PR #1989, opened 2026-09-07
Fix route InfoBlock save and download dialogs

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1977

alisa911:
## AI disclaimer

1. Study issue #1977, find the bug, make the fix. The agent traced `selectedGpxFile.save` to `PanelButtons`, which since `c47cbd6b1` renders only for local tracks, and rendered `SaveTrackDialog` in `RouteTrackActionsButtons` the way `TravelTrackActionsButtons` already does.
2. Are you sure about the fix? Why does this dialog appear when I press Edit? — Explained: the stale `save` flag is copied into the local track on Edit and only then hits `PanelButtons`; the fix closes that path.
3. Why doesn't any selenium test fail, do we have no tests for this? — Checked: test 42 only lists the button ids, test 43 saves through the Navigation header button, which uses `SaveTrackDialog` with an external track and never touched the broken code.
4. Add the test. Extended `43-navigation-save-download.mjs` with save through the route InfoBlock.
5. Add a check for all three buttons in the row: order via `enumerateIds`, Download opens its dialog, Edit turns the route into a local track (last, since it changes the object type).
6. The run failed on `se-download-track-full` from the InfoBlock. Identified as a second bug — `DownloadTrackDialog` got no `navTrack`, so it auto-downloaded without a dialog — and asked whether it is a bug or intended before touching code. Confirmed as a bug, fixed with `navTrack={track}`, as the header button does.
7. Commit name, then a shorter one.

No commits were made by the agent; the test was run by the author locally.

Issue #1065, opened 2025-10-03
Search: it´s not possible to search favorites or tracks by the name


Rowin63:
The search option (magnifying glass) does not search inside the users data, both search switches seem to search POI only.
It´s not possible to search and find a certain favorite or track by it´s name.
Or this option is hidden very well and I don´t find it.

![Image](https://github.com/user-attachments/assets/7f00230a-604f-41d0-acc0-514b66ce658a)

Both open this POI search

![Image](https://github.com/user-attachments/assets/52f4763b-6662-4b20-89cb-7d6241146068)

vshcherb:
We have a straight forward approach to do this as we implemented in Android to search with criteria StringMatcherMode.CHECK_STARTS_FROM_SPACE

Zirochkabila:
# Testing 1

> Web - prod

Tested display of both Tracks and Favorites in search results (if they are shown on the map but not displayed) - **Fixed**

<details><summary>Details</summary>
<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/8cac5352-9ec3-419b-b767-9fb51dda7fb5"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/421d66e1-dc8d-4775-955f-09d57383c095"  width="300" height="auto"/> </th>
  </tr>
</table>  
</details>

dmpr0:
## UI Review

<table style="width:100%">
  <tr>
    <th>Tracks</th>
    <th>Search results</th>
    <th>Android</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/e40d8970-f06b-4b7c-b81c-9d8b1ef56918"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/3b17a04d-cc13-4792-8b18-5837fc4b5784"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/55efff11-b51a-487d-88fc-4652fe3349d2"  width="300" height="auto"/> </td>
  </tr>
</table>  

- [ ] Search for GPX is not working

Zirochkabila:
# Testing 2

> Web - prod

**Severity**: 🟡 Minor / **Priority**: 🟠 Normal

* [ ] 1. Searching for tracks by name works, but the order of the search results is unclear: the track centered on the map appears 7th in the results list.

<details><summary>Details</summary>
[...]

PR #1990, opened 2026-09-07
Add unit tests for search

alisa911:


alisa911:
https://github.com/osmandapp/web/issues/1065

alisa911:
## AI disclaimer

Written with Claude (Cowork, Claude Opus 5), driven by [@alisa911](https://github.com/alisa911). Summary of the requests it worked from, in order:

1. Asked which unit tests are still worth adding for search, and then for the poi context menu, the map markers and anything else found. The agent read the search surface and proposed five areas; `OpeningHoursInfo` was dropped after checking that the server sends ready strings and the component only splits them.

2. Start with poi types. Covered `isBrandType` / `parseBrandType`, `getTopIndexValueName`, `getCategoryName`, `formattingPoiType` and the icon fallback chain of `getIconNameForPoiType` / `getCatPoiIconName` in `search/00-poi-types.test.js`. No bugs found; the `iconName !== 'null'` string check was confirmed as a guard against the literal `"null"` from the server and pinned by a test.

3. Continue with the search result row. The agent stopped before writing code: `searchTypeMap` and the result property names lived in `map/layers/SearchLayer.js`, which cannot be loaded in jest (react-leaflet is ESM, the layer pulls in leaflet.markercluster), so with the stub every branch compared `undefined` to `undefined` and the tests would have proved nothing. Asked whether to move the constants; the author said yes. They now live in `manager/searchConstants.js`, `getObjIdSearch` moved to `SearchManager`, imports updated in 11 files, and no manager or menu imports the map layer any more.

4. Cover the results and the features built from them. `search/01-search-results.test.js` covers `getPropsFromSearchResultItem` (six kinds of result, empty poi name, address lines, the street case), `getPoiParentCategory`, `searchByCategory` and `buildSearchParamsFromQuery`; `search/02-search-features.test.js` covers `buildFavoriteFeatures` and `buildWptFeatures` (own vs shared file with the same path, a waypoint that moved, appearance from the waypoint group), `getObjIdSearch`, `parseTagWithLang` and `getPhotoTitle`.

5. Asked why no bugs were reported, then to fix the one the agent had worked around. Loading the real `MarkerOptions` in jest crashed: `FavoritesManager` built its list of marker shapes at module level from `MarkerOptions`, which imports it back through the poi managers, so in that import order the shapes were `[undefined, undefined, undefined]`. Fixed by building the list on call (`getShapes()`), and the unused `shapes` field was dropped from the module export.

6. Asked to untangle the cycle itself instead of living with it. `getFirstSubstring`, `preparedType` and `parseTagWithLang` moved out of the search menu component into `PoiManager`, `SEARCH_BRAND` into `searchConstants`, and the `getFirstSubstring` parameter of `getCategoryName` disappeared - it existed only to avoid that import. Then the same pattern was fixed everywhere else: four `store/geoRouter` files and two favorite components read shared constants through imports instead of module-level copies, and the three wpt shape constants became named exports. A scan of all 449 source files reports no module-level reads left.

7. Cover the poi context menu. Three bugs in `WptTagsProvider`: `isUrl` accepted any scheme, so the recommended `wikipedia=de:München` tag was taken for a link and rendered as a dead `href="de:München"`; the article link was never built at all, because the call to `addWikipediaTags` had been removed in a36f5a1e1 while the leftover `getWikipediaURL` result was overwritten by the formatting below it; and `parseUrl` matched a language code as `[a-z]+`, so `be-tarask`, `zh-min-nan` and the like never opened. Also `facebook=facebook.com/name` produced `https://facebook.com/facebook.com/name`. All fixed, dead code removed, covered by `poi/00-poi-tags.test.js`.

8. Cover the markers. Two bugs in `changeIconSizeWpt`: `width="(\d+)"` had no word boundary, so it matched `stroke-width` and set the stroke to the size of the shape, and the x-axis rotation of an arc was scaled as if it were a length. Fixed and covered in `markers/00-icon-svg.test.js`, together with the color helpers, the icon path fallbacks and `resolveWptAppearance`.

9. Said the suite had grown too large. The new files were compressed from 109 tests to 56 by grouping the cases of one function into one test and dropping tests of one-line guards and of generated resource content; the mutation check was rerun afterwards, each fix still fails its own test. The rules that produced this were written into the `add-test` skill, together with "never assert against a stub" and what to look for while writing tests.

10. Three Copilot review comments were accepted and fixed: the wikipedia article was still opened from the raw tag value (`getWikipediaContent` now reads `tag.url ?? tag.value`), the arc parameter counter reset after the first flag pair so a second implicit set inside one `A` had its rotation and flags scaled, and a duplicated word in a comment.

No commits were made by the agent; every commit was reviewed and made by the author. The 334 unit tests were run by the agent, the selenium suite by the author.

PR #1992, opened 2026-09-08
Add unit tests

alisa911:
## AI disclaimer

Written with Claude (Cowork, Claude Opus 5), driven by [@alisa911](https://github.com/alisa911). Summary of the requests it worked from, in order:

1. Cover the router profiles and their parameters, then "can it be fewer tests?". `routing/00-router-profiles.test.js` covers `getProfile` and its fallback when a router or a profile is gone, the profile stored on a track point and its short form, the params of a profile and what counts as a change, and that the url of a profile wins over the url of its provider. One oddity found and not fixed: `geoRouter.getURL` (`getters.js:153`) is the only getter without the provider fallback, unreachable from the current callers - recorded in KNOWN_ISSUES.md instead.

2. Cover the rules of the track editor, again asking for fewer tests, and asking whether the add-test skill was used at all. Eight tests in `tracks/12-track-editor.test.js`: which segments the editor may change, `splitProtectedSegment`, and the points that are sent to the router. Grouped per function from the start after that.

3. Cover local tracks. The agent first called a delete path an urgent data-loss bug without checking its callers; the author tested it by hand and saw nothing, and the agent had to admit the path is only reachable from the nameless-track cleanup. Writing the tests then exposed a second, user-reachable bug: `deleteLocalTrack` deleted by an index taken from the in-memory list, and `deleteLocalTracksByIndexes` removed the records it had just written, so a track deleted in one session came back after a reload. Both fixed in `LocalTrackStorage.js`, four tests in `tracks/13-local-tracks.test.js`.

4. Cover marker clustering. Three findings in `Clusterizer.js`: the poi branch introduced in 5daeb7bf6 hardcoded 64/12 px radii and stopped reading `iconSize`, so the travel points that `TravelLayer` had asked to keep 8/4 px apart were thinned to 64 px and the closer ones dropped; explore allowed 21 big markers instead of 20 (`<=`); and the explore branch sorted the array of the caller in place. Fixed with `mainRadiusPx` / `secondaryRadiusPx`, covered by five tests in `markers/01-clusterizer.test.js`. `getPointLatLon` moved from `TrackLayerProvider` to `Clusterizer` (its only user) so the module can be loaded in jest without react-leaflet.

5. Cover weather. `WeatherManager.dayFormatter` built the day of the forecast key from the local date while `timeFormatter` used UTC hours, and the server keys those rows in UTC (`WeatherController`, `SimpleDateFormat("yyyyMMdd_HH")` with the UTC time zone), so for every user whose local date differs from UTC the row was not found or the row of the neighbouring day was shown. Fixed, and the duplicate UTC-correct copy of the same function in `useWeatherLocationChange` removed. Five tests in `weather/00-weather.test.js`.

6. Asked what is left to cover in favorites and whether tests are needed there at all; the agent proposed a list and the author picked it up. Six tests in `favorites/00-fav-groups.test.js` (group name to file name and back, a free name for a new group, hidden and pinned restored from the IndexedDB cache, the menu list with distances, the size of a group). A suspected bug - the escaped group name passed to `addHidden` - was disproved by a mutation check and reported as not a bug; `isHidden` reading `point.ext.extensions.hidden` without optional chaining was guarded, as everywhere else in the project.

7. Cover the rebuild of the groups after a favorite changes. Four tests in `favorites/01-fav-changes.test.js` over `prepareResult`, `updateGroupAfterChange` and `updateSelectedGroup`. The second suspicion from the previous item - `updateFavFile` copying every key of the response into `group.file` - was checked against the server's `WebGpxParser.TrackData`, which has no `name` field, so the file name and the group id are not affected.

8. Two rounds of review comments applied. From Copilot: `getPointLatLon` unexported, a stale comment in the jest config, and the poi branch of `clusterMarkers` moved above the distance calculations it never used - which let the now-dead `isPoi` parameters go from `calculateDistance`, `getMaxMainPlaces`, `getUseUniformMarkerPlacement` and `createMainMarkersArr`. From the review of this PR, seven points, all confirmed against the code: a test in `01-fav-changes.test.js` pinned broken behaviour - `FavoriteHelper.updateSelectedGroup` builds the file name without the `favorites-` prefix and the id without `userid`, so a group added by that branch can never match the one the server returns - the assertions were dropped and the bug recorded in KNOWN_ISSUES.md; an orphaned stub and a no-op mapping removed from the jest setup; `fakeTracksDb` renamed to `fakeIndexedDbStore` since it also backs the favorites cache; comments restored and updated; and `iconSize` dropped from the three poi callers where it is no longer read.

No commits were made by the agent; every commit was reviewed and made by the author. The 373 unit tests were run by the agent, the selenium suite by the author.

### DIFF
--- a/map/src/context/LocalTrackStorage.js
+++ b/map/src/context/LocalTrackStorage.js
@@ -181,10 +181,7 @@ export function deleteLocalTrack(ctx, index = -1) {
     const currentTrackIndex =
         index !== -1 ? index : ctx.localTracks.findIndex((t) => t?.name === ctx.selectedGpxFile?.name);
     if (currentTrackIndex !== -1) {
-        deleteTrackFromDB(currentTrackIndex).then(() => {
-            ctx.localTracks.splice(currentTrackIndex, 1);
-            ctx.setLocalTracks([...ctx.localTracks]);
-        });
+        deleteLocalTracksByIndexes(ctx, [currentTrackIndex]);
     }
 }
 
@@ -221,12 +218,8 @@ export function deleteLocalTracksByIndexes(ctx, indexes = []) {
                     }
                 }
 
-                // collect all ids to remove: deleted + moved-from
-                const idsToRemove = new Set([...toDelete, ...moves.map(({ oldId }) => oldId)]);
-
-                // delete old ids
-                for (const oldId of idsToRemove) {
-                    await deleteTrackFromDB(oldId);
+                for (let id = next.length; id < prev.length; id++) {
+                    await deleteTrackFromDB(id);
                 }
             } catch (e) {
                 DEBUG && console.error('Failed to delete selected local tracks', e);
--- a/map/src/context/LocalTrackStorage.js
+++ b/map/src/context/LocalTrackStorage.js
@@ -218,6 +218,7 @@ export function deleteLocalTracksByIndexes(ctx, indexes = []) {
                     }
                 }
 
+                // the ids past the new length are free after the tracks left were saved over the first ones
                 for (let id = next.length; id < prev.length; id++) {
                     await deleteTrackFromDB(id);
                 }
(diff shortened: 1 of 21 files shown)


alisa911:

PR #1993, opened 2026-09-08
Add unit tests

alisa911:
## AI disclaimer

1. After a review of the previous PR found a test that pinned broken behaviour, asked to go through every unit test again and check whether each expectation is what the app should do, not what the code does. All 31 files were checked against the server (osmandapp/tools) and Android (osmandapp/android). One test was found to pin a real divergence - the web escapes both `/` and `:` in a favorite group file name while Android escapes only the first one it finds - and a few more doubtful cases; they went into KNOWN_ISSUES.md (not part of this PR), and the rule "expect what the app should do, not what it does" went into the add-test skill.

2. Cover the routing cache of the track editor. `effectControlRouterRequests` let a seventh request through when six were in flight (`>` instead of `>=` against `MAX_STARTED_ROUTER_JOBS`); fixed after asking. Six tests in `routing/01-routing-cache.test.js`: a segment is cached by its points and profile, a failed request is not repeated, a routed segment goes into the track once, undo restores the geometry from the cache, the debouncer keeps the last call.

3. Cover the distances of a track. `getTrackPoints` shifted `distanceTotal` of the second and following `<trk>` of a file in place, and it is called on every redraw of the graph, so the distance axis grew on each call. The author asked to re-check that the bug is real before fixing; the chain was verified end to end (`WebGpxParser.getTracks` returns one entry per `<trk>`, `addDistance` counts each from zero, `getAllPoints` returns references, `GpxGraphProvider` reads `distanceTotal`). Fixed by shifting copies; seven tests in `tracks/14-track-distance.test.js`. The first version of the test compared mutated points with themselves and passed - caught and rewritten to snapshot the numbers first.

4. Cover deleting and reordering the points of a track. A point dragged to the first place in the Points tab kept the geometry of its old segment, so `validateRoutePoints` rejected the track and the analysis and the save ignored the reorder; fixed in `insertPoint`. The author reminded about the add-test skill: the inline contexts of the new tests were moved into `fixtures/tracks.js` (`createEditorCtx`, `createEditorPoint`, `createRoutedPoints`). Five tests in `tracks/15-track-points.test.js`.

5. Cover the conversion of a navigation route into a track. No bugs; three tests in `navigation/00-route-to-track.test.js` (segments cut at the route points, a via point snapped to the line, alternatives left out, elevation borrowed at the ends of the line). Two doubts noted without a fix: a round trip whose finish is on the start collapses both into one map entry, and the point shared by two segments was counted twice in the average elevation - the latter was fixed on review, see 8.

6. Cover the cloud changes and trash. `deleteVersionsFromMenu` kept the header of a month whose files were all deleted as long as any later month still had a file; the author asked to re-check, the rendering in `CloudChanges` / `CloudTrash` was confirmed to draw headers as they come; fixed. Five tests in `settings/00-cloud-changes.test.js`.

7. Cover the url helpers, the file storages and the matched objects of the spatial search. No bugs; seven tests in `util/07-url-and-storage.test.js` and `search/03-matched-objects.test.js`. A matched amenity without coordinates would be listed as a link that does nothing - not pinned by a test, since the server builds matched objects from map objects with coordinates.

8. Four review comments applied: `deleteFile` now goes through `deleteVersionsFromMenu` like `restoreFile` (the test had pinned the leftover month header again); `isLast` of the row that became the last of its month is recomputed, so no divider is drawn before the next header; the point shared by two segments of a route counts once in `createAnalysisFromRoute` (the test expects 115 with the arithmetic written down); the skill names the repositories instead of local paths.

No commits were made by the agent; every commit was reviewed and made by the author. The 406 unit tests were run by the agent, the selenium suite by the author.

Screenshot 2026-09-08 at 15 51 53

### DIFF
--- a/.claude/skills/add-test/SKILL.md
+++ b/.claude/skills/add-test/SKILL.md
@@ -69,6 +69,16 @@ A test earns its place only if the behaviour can break silently and the user wou
 
 A manager or a formatting module is 10-15 tests, not 30.
 
+## Expect what the app should do, not what it does
+
+The expected value comes from the contract, never from running the code and copying the result. Before pinning a value, find where it is decided and check there:
+
+- a file name, a key, a format shared with the phone or the server - read the Android code (`~/osmand/android`, e.g. `FavouritesFileHelper`) or the server (`~/osmand/tools`, e.g. `WebGpxParser`, `MapApiController`) and expect what they produce; a test that pins the web's own encoding proves nothing when the other side disagrees;
+- a value shown to the user - ask whether a person would call it right (a date order, a name with the extension, a lowercase language);
+- a limit - check what it is measured against (255 bytes of the name, or of the name with the prefix and the extension on the phone).
+
+If a suspected bug's test passes on the first run, suspect the test: it compares live references with themselves (snapshot the numbers before the second call), or the fixture is not what the app really holds (`favorites/favorites.gpx` is not a favorite file name). Write down the reason a value is expected when it is not obvious from the name of the test.
+
 ## Look for bugs while writing
 
 Writing the test is the review. Report the bugs first, before describing the tests, and never write a test that blesses broken behaviour.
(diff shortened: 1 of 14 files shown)


alisa911:

PR #1994, opened 2026-09-09
Speed up selenium tests

alisa911:
Full run: 3028s → 1702s (−44%), 78/78 passed.
## AI disclaimer

1. Study the selenium suite and find out whether it can be sped up without changing what the tests check. Static reading of `lib.mjs`, `options.mjs`, actions and all 78 tests pointed at the test infrastructure, not the tests: a 1s implicit wait mixed with explicit waits, fixed 550 ms pauses around `MenuItem`/`Select` clicks and a CSS-transition pause before every MUI click, a dead `se-show-main-menu` id waited for in 42 places, a 3s `FINAL_DELAY`, and `performance` logging collected in every test but read only with `--verbose`.
2. Run the tests instead of estimating. The suite was run in the agent's container against test.osmand.net with a per-category profiler: baseline 3395s, safe changes 2839s, all changes 2416s (-29%). The profile confirmed the implicit wait (~1050s in empty `findElements`) and the fixed click pauses (~700s) as the main cost. Test 13 exposed a real race in `actionImportCloudTrack` (a single upload opens the track, the test clicked back too early); fixed by waiting for `se-track-context-menu`.
3. Apply the changes to the repo, no commit. `options.mjs`: `implicit: 0`, performance log only with `--verbose`, `FINAL_DELAY` 1s; 42 `se-show-main-menu` clicks removed; `lib.mjs`: fixed pauses replaced with a wait for running layout transitions (`document.getAnimations()`).
4. Almost all favorites tests fail locally with "element not interactable: has no size and location" - how could this be sent. The agent's first attempt (an element-ready check over two animation frames) was rejected by the author as over-engineered and did not fix it. The failure was then reproduced on a dev server started in the container: after login the app navigates on its own and the next menu click raced it; separately, menu panels are always mounted and `display:none`, so `waitBy` returned buttons of hidden panels. Fixed with `actionIdleWait()` at the end of `actionLogIn` and a `getRect()` size check in `clickBy` that retries until the element is laid out; the animation wait was cut down to 12 lines. Favorites 16/16 in the container.
5. Fix the remaining failing tests one by one (87, 85, 25, 96), running only those. 87: `actionIdleWait()` after typing before triggering the exit guard, the "temp fix" with three `idleWait(3000)` replaced by waiting for the dialog and re-triggering if the map pan dismissed it. 85: cleanup via `actionDeleteAllFavorites` so leftovers from other tests do not break the empty-page check. 25: leftover `se-test-create-track` from a failed run made save show "update the track?"; deleted on start. 96: `deleteFileVersion`/`restoreFile` do not refresh the list, leaving Changes triggers `list-files` before the server applied the change; `actionIdleWait()` after both actions. 71/79/43 got the same wait after save-to-cloud. `clickBy` timeouts now name the locator.
6. Explain the "Cannot convert undefined or null to object" crash seen in 25 and on prod. Traced to `Object.keys(savedVisible)` in `GlobalFrame.processTracks`; the guard from the author's own commit `f8918243b` is present in HEAD and in both deployed bundles, so the crashing bundle was stale.
7. More ideas, code untouched: second worker on a second account, single-check optional lookups, tiles wait only for map actions, bundle cache between tests, API login. The author chose two of them: `waitBy` got `now` (implied by `idle`) for one-shot optional checks, `actionIdleWait` got `tiles` and waits for tiles only in map helpers, `actionOpenMap` and `actionFinish`. The second worker was analysed (`TEST_LOGIN2` is a fixture for 09 and the second user of 99, so it cannot be cleaned) and postponed.
8. Update the `add-test` skill with what was learned about selenium waits, app races and cleanup.

All test runs used for numbers were made by the agent in its container against test.osmand.net and a local dev server; the final state was run by the author locally. No commits were made by the agent; the author committed and named the commits.

**Results:**
Screenshot 2026-09-08 at 15 51 53
Screenshot 2026-09-09 at 17 05 55

Screenshot 2026-09-09 at 17 17 44


alisa911:

PR #1996, opened 2026-09-09
Add MVT map objects context menu

alisa911:
## AI disclaimer

1. Study osmandapp/OsmAnd-Issues#3263, the web and the server projects and write a plan; the Dev notes are recommendations, the Expected Result (tags equal to search results and the mobile app) is binding. Follow-ups: no raw tile tags in the preview, click only on what Android considers clickable. `MvtLayer` got `handleMapClick` and a pointer cursor over clickable symbols, `MvtObjectSelection.pickClickableFeature` accepts symbol layers without line placement and with `osm_id` (as `MapSelectionHelper`), the devel tag popup was removed, the full object is loaded by `WptDetails.loadMvtObject`.
2. Clicking a house number showed "Residential area" and "Poi undefined": show the address card as the address search does. `createSearchObjectOptions(HOUSE)` is reused; a hardcoded tag list the agent had added to guess the type was rejected and removed, type and icon come only from the server; `SearchResultItem` no longer renders a type for a POI without one; a favorite added from an address had a broken icon, fixed with `FINAL_POI_ICON_NAME` and the default wpt icon as in Android.
3. Behave like the rest of the web: update the URL (`navigateToPoi`/`navigateToStop` with `replace`, the previous object's URL must not stay), no selected pin for the MVT click, no extra stop marker when a stop is clicked (guard in the `TransportStopsLayer` `stopByUrl` effect), no address flicker, the name must not jump while the type is unknown (NBSP type placeholder and a 48 px icon placeholder), no card under the loading indicator, the duplicated "Housenumber" tag row removed. The author kept the preview and asked for loading indicators only after 1 s (`useDelayedFlag`); a double flicker on a stop click was traced to a second fetch and removed.
4. Compare with Android what is missing. Visible tags go through the server as for search; the hardcoded transport stop tags and the osm type calculation were moved to the server (`web_transport_stop_id`, `get-poi-by-osmid` without `type`), then, after a parking click showed an empty header, objects outside the POI index are loaded from their tile tags. Multiple objects under the click were discussed and dropped: 5 px tolerance, symbols do not overlap, no touch support on the web.
5. Several review passes for duplication and simplification: `createMapObject`, `stringifyTagValues`, `createSearchObjectOptions`, `getFinalPoiIconName` and the `SearchApi` wrappers replaced copies in `PoiLayer`, `SearchLayer`, `Clusterizer`, `SearchResults`, `TransportStopsLayer` and `WptDetails`.
6. Unit tests only where needed, then a selenium test for the vector tiles (NOVUS header, tags and four buttons, a stop with Routes, a house), written with the project `add-test` skill after the author pointed out it exists and must not be edited: two dense tests in `00-mvt-object-selection.test.js`, a no-type case in `01-search-results.test.js`, `38-mvt-context-menu.mjs`. Test runs by the author exposed: no WebGL in headless Chrome (software GL flags in `options.mjs`), the info block closes with the favorite panel, navigation left open (closed by the header button), a page reload per click (hash navigation instead), a click before the symbols were placed (`seIsTilesLoaded` only when MapLibre reports the tiles loaded), and a regression in `29-transport-stops.mjs` fixed by setting the object type before the selected stop in the marker click.
7. Review of the PR: `loadMvtObject` makes one `get-poi-by-map-object` request instead of `get-poi-by-osmid` and `get-poi-by-tags`. Kept with reasons: the preview address is resolved for the clicked point, the 1 s spinner delay applies to all objects as requested, and a `.catch` suggested by Copilot is not needed because `apiGet`/`apiPost` do not reject without `throwErrors`.

The agent made no commits. The unit tests were run by the agent, and each was checked to fail with the code broken. The selenium test was run by the author on the local server and by the agent in its own environment against a dev server proxied to test.osmand.net; the final version after the last test fixes was run by the author.

Screenshot 2026-09-10 at 14 12 32

### DIFF
--- a/map/src/infoblock/components/wpt/WptDetails.jsx
+++ b/map/src/infoblock/components/wpt/WptDetails.jsx
@@ -62,6 +62,8 @@ import WptTagsProvider, {
     addWikidataTags,
     getOsmIdFromOsmUrl,
     POI_ID,
+    TRANSPORT_STOP_ID,
+    stringifyTagValues,
 } from './WptTagsProvider';
 import WptTagInfo from './WptTagInfo';
 import { useTranslation } from 'react-i18next';
@@ -102,13 +104,19 @@ import {
     liveHash,
 } from '../../../manager/GlobalManager';
 import { useWindowSize } from '../../../util/hooks/useWindowSize';
+import { useDelayedFlag } from '../../../util/hooks/useDelayedFlag';
 import gStyles from '../../../menu/gstylesmenu.module.css';
 import { buildSearchParamsFromQuery } from '../../../util/hooks/search/useSearchNav';
-import { isFavoriteFromSearch, isWptFromSearch, navigateBackToSearchResults } from '../../../manager/SearchManager';
+import {
+    createMapObject,
+    isFavoriteFromSearch,
+    isWptFromSearch,
+    navigateBackToSearchResults,
+} from '../../../manager/SearchManager';
 import { useLocation, useNavigate } from 'react-router-dom';
 import LocationInfoLine from '../common/LocationInfoLine';
 import OpeningHoursInfo, { getOpeningHours } from './OpeningHoursInfo';
-import { getPoiByOsmIdApi, getTransportStopApi } from '../../../manager/SearchApi';
+import { getPoiByOsmIdApi, getPoiByTagsApi, getTransportStopApi } from '../../../manager/SearchApi';
 
 export const WptIcon = ({ wpt = null, color, background, icon, iconSize, shieldSize, ctx }) => {
     const [iconState, setIconState] = useState({ svg: null, isLoading: true });
@@ -172,9 +180,10 @@ export function getObjType(wpt, t) {
 export const ADDRESS_NOT_FOUND = i18n.t('web:no_data');
 export const TYPE_NOT_FOUND = 'No type';
 export const EMPTY_STRING = '';
-const ADDRESS_TAG_PREFIX = 'addr:';
 // keeps the header layout of an MVT preview until the type and icon are loaded
 const TYPE_PLACEHOLDER = '\u00A0';
+// loading indicators are shown only for a long loading, not to flicker
+const LOADING_DELAY_MS = 1000;
 
 export async function getAddressByLatLon(lat, lon) {
     if (lat == null || lon == null) return null;
@@ -213,6 +222,9 @@ export default function WptDetails({ setOpenWptTab, setShowInfoBlock }) {
 
     const [isAddressAdded, setIsAddressAdded] = useState(false);
     const [isPhotosAdded, setIsPhotosAdded] = useState(false);
+
+    const showLoading = useDelayedFlag(loading, LOADING_DELAY_MS);
+    const showPendingLoading = useDelayedFlag(Boolean(wpt?.pending), LOADING_DELAY_MS);
 
     const renderedTags = useMemo(() => {
         const tags = wpt?.tags?.res;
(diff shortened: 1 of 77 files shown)


alisa911:

RZR-UA:
Some objects have neither tags nor icons.

Additionally, please reduce the precision for `pin=lat,lon` (6 digits are enough).

Інститут проблем сучасного мистецтва
18-Д вулиця Євгена Коновальця, Київ
Location: geo:50.42446408954853,30.5281455561845?z=18
https://test.osmand.net/map/poi/?pin=50.42446408954853,30.5281455561845

Авіаційно-космічний центр КПДЮ
7а Фортечний тупик, Київ
Location: geo:50.42449483759908,30.53058578070955?z=18
https://test.osmand.net/map/poi/?pin=50.42449483759908,30.53058578070955

Новопечерський квартал №5
7Б Фортечний тупик, Київ
Location: geo:50.423176069942286,30.530564328186234?z=18
https://test.osmand.net/map/poi/?pin=50.423176069942286,30.530564328186234

map/src/infoblock/components/wpt/WptDetails.jsx
```diff
@@ -167,6 +180,10 @@ export function getObjType(wpt, t) {
 export const ADDRESS_NOT_FOUND = i18n.t('web:no_data');
 export const TYPE_NOT_FOUND = 'No type';
 export const EMPTY_STRING = '';
+// keeps the header layout of an MVT preview until the type and icon are loaded
+const TYPE_PLACEHOLDER = '\u00A0';
```
RZR-UA, line 184: What a magic const 00A0? Try `&nbsp;`?

map/src/map/layers/MvtLayer.js
```diff
@@ -28,7 +34,7 @@ function getPublicAssetUrl(path) {
 }
 
 function createStyle(baseStyle, tileUrl, options = {}) {
-    const style = JSON.parse(JSON.stringify(baseStyle));
+    const style = structuredClone(baseStyle);
```
RZR-UA, line 37:
```
export function copyObj(obj) {
    return typeof obj === 'object' ? JSON.parse(JSON.stringify(obj)) : obj;
}
```

Please apply structuredClone() to this method instead of the JSON-to-JSON approach.

tests/selenium/src/options.mjs
```diff
@@ -193,7 +193,9 @@ export async function prepareDriver() {
 
     mobile && options.setMobileEmulation({ deviceName });
     // headless && options.headless().windowSize({ width, height }); // use --window-size
-    headless && options.addArguments('--headless', '--disable-gpu', '--no-sandbox', `--window-size=${width},${height}`);
+    headless && options.addArguments('--headless', '--no-sandbox', `--window-size=${width},${height}`);
+    // software WebGL for the MVT (MapLibre) layer
+    headless && options.addArguments('--use-gl=angle', '--use-angle=swiftshader', '--enable-unsafe-swiftshader');
```
RZR-UA, line 198: Before merging, please try the options without `--disable-gpu` on the builder. There were complications.

RZR-UA (approved):

map/src/infoblock/components/wpt/WptDetails.jsx
alisa911: `&nbsp;` simply cannot work here. The constant does not go into markup, it goes into the JS prop `name`

map/src/map/layers/MvtLayer.js
alisa911: `copyObj` is used only by the routing store and is out of scope here. Switching it would also break the unit tests: `structuredClone` is undefined in jsdom 20, so it needs a polyfill in the jest setup. Better as a separate change.

Issue #2014, opened 2026-09-15
Shared wiki POI link: wrong map position + unrelated photos


alisa911:
https://osmand.net/map/poi/?type=osmwiki&wikidataId=11663091&pin=35.3,138.7

- [x]  1. Map is not centered on the pin

The shared POI link has no map position in it, so the map opens at the reader's own location instead of the POI —
the pin is somewhere off-screen.

Expected: `pin` defines the map center, as in Android.

- [ ] 2. "Online photos" shows Russian propaganda images

Out of 9 photos 8 are staged photos of Vladimir Putin — the head of the state waging the war against Ukraine,
wanted by the ICC. Showing them to our users is unacceptable, no matter which Commons category they come from
(here: https://commons.wikimedia.org/wiki/Category:2011_Shizuoka_earthquake).
Expected: no such photos in the app.

<img width="419" height="561" alt="Image" src="https://github.com/user-attachments/assets/7075a720-452f-4003-8d0b-5517b6dfa3d0" />

PR #2015, opened 2026-09-15
Center the map by the URL pin

alisa911:
A link shared from a POI (`/map/poi/?...&pin=lat,lon`) carries no map position, so the map opened at the reader's own geo-IP location and the pin stayed off-screen. `pin` has never defined the map view — since it was introduced it only drew a marker, the view always came from the hash.

Now `pin` defines the initial center whenever it is present: zoom is taken from the hash if there is one, otherwise the default 9 (the zoom used when the map centers by geo-IP). The hash is rewritten from the pin before the map is created, so leaflet-hash does not pull the map back and geo-IP detection stays off.

Same semantics as Android, where `pin` is the point and the hash gives only the zoom (`GeoPointParserUtil.parseSimpleDomainsUri`, `IntentHelper.parseAmenity` / `showAmenity`).

Checked locally against test.osmand.net:

| URL | Result |
|---|---|
| `poi/?...&pin=35.3,138.7` | center 35.3/138.7, zoom 9, hash built from the pin |
| `poi/?...&pin=35.3,138.7#12/50.4584/30.5310` | center 35.3/138.7, zoom 12 from the hash |
| `?pin=48.8584,2.2945` | center on the pin, pin marker shown |
| `#12/48.8584/2.2945` (no pin) | unchanged |
| `/map/` (no pin, no hash) | unchanged |

Covered by `tests/selenium/src/tests/map/88-pin-url-center.mjs`.

## AI disclaimer

Written with Claude Code (Claude Opus 5), driven by [@alisa911](https://github.com/alisa911). Summary of the requests it worked from, in order:

1. Check whether a wiki POI link with a `pin` and a hash pointing elsewhere is a bug or intended. Reproduced on osmand.net: the POI marker is added at the pin, the map stays where the hash points; the `pin` query parameter never affected the map view.
2. Correct the diagnosis: the bug is in the shared link, which has no coordinates, and the map jumps to the reader's location. Confirmed: `sharePoi` in `PoiActionsButtons.jsx` builds the URL from pathname and search only, and with an empty hash `detectGeoByIp` flies the map to the geo-IP location.
3. Ask whether this was a regression or the intended logic. History check: `pin` only ever drew a marker (since 2022), while the hash was dropped from the shared POI link in `78fa9433e` "remove hash from poi share link" (PR #1157, 28.10.2025) — before it the link was shared with the hash.
4. Make `pin` always define the center and build the hash from it when there is none, on a new branch. Added `getInitialView` in `OsmAndMap.jsx`, extracted `getPinPointFromUrl` in `MapContext.js` so the pin is parsed in one place, and exported `flyZoom` for the default zoom.
5. Ask whether the Android behaviour was checked. It had not been: `GeoPointParserUtil.parseSimpleDomainsUri` takes the point from `pin` and only the zoom from the fragment, and `IntentHelper.showAmenity` centers on the pin with the last known zoom. The implementation matches; the only difference is the default zoom, since the web has no stored last zoom.
6. Add a selenium test for the different pin cases. Added `88-pin-url-center.mjs` with four cases. Reverting the fix showed that the "pin wins over the hash" case passed anyway on `/map/poi/`, where `panToIfNeeded` in `PoiLayer` pulls the map to the loaded POI, so that case was moved to `/map/`, where only the new code decides the view.
7. Open the pull request and split the change into several commits.

Test runs (selenium 88, the whole map category, `90-poi-url`, `29-transport-stops`, 407 unit tests, lint and prettier) were done by the agent against a local dev server; the review and the merge are not.


alisa911:
https://github.com/osmandapp/web/issues/2014

PR #2038, opened 2026-09-17
Fix MVT style map jittery on wheel/button zoom in/out (also fix related regressions)

alisa911:
Wheel zoom on the map is now fractional (hundredths of a level) and eased frame by frame, so on the MVT style MapLibre renders every frame itself instead of Leaflet scaling the canvas bitmap to the next whole level and snapping to the native render at the end.

- `MapStateLayer`: own wheel handler instead of Leaflet `scrollWheelZoom`, raw `deltaY / 450` per level as MapLibre's `wheelZoomRate`, eased through `map._move` pinch frames, finished with `_move` + `_moveEnd` (no `viewreset`, raster tiles are kept).
- `OsmAndMap`: `zoomSnap 0.01`, `updateWhenZooming false` on the raster layer.
- `util/mapHash`: leaflet-hash writes the zoom with two decimals and reads it with `parseFloat`; `initialMapView` too.
- `PoiLayer`: `search-poi` keeps getting an integer zoom (`Integer` on the server).
- `MvtLayer`: default MapLibre tile fade instead of `fadeDuration: 0`.

Measured on the raster style, one wheel notch: 12 tile requests and the same event sequence as Leaflet's own wheel zoom (zoomstart 1, zoomend 1, moveend 1, no viewreset).

## Video

[final.webm](https://github.com/user-attachments/assets/57e4c48b-6af8-465e-9065-61fb96bb7b67)

## AI disclaimer

1. Why do vector tiles on osmand.net look jerky when zooming, while MapTiler is smooth. Finding: MapTiler is driven by MapLibre itself, our map is driven by Leaflet through `@maplibre/maplibre-gl-leaflet`, which CSS-scales the GL canvas bitmap for the 250 ms zoom animation and only calls `jumpTo` on `zoomend`; the zoom steps are whole levels (Leaflet defaults) and `MvtLayer` passed `fadeDuration: 0`.
2. Publish the analysis as an artifact with a video of the current behaviour, then a cheap fix and a real fix, each with a video. Recorded with headless Chrome via CDP screencast (same wheel input in every recording); fix 1 was `zoomSnap 0.25` plus default fade, fix 2 a custom wheel handler with fractional zoom eased frame by frame so MapLibre renders every frame.
3. Pushback: the URL showed `#16.745795355902786`. The hash now keeps two decimals and is parsed with `parseFloat` (leaflet-hash and `initialMapView`).
4. Record a whole-level variant in case fractional zoom is rejected, then "do fix 3". The self-review of that variant found that ending the animation through `setZoomAround` fires `viewreset`, which made `GridLayer` reload all raster tiles (30 tile requests per wheel notch against 12 on main); replaced by `_move` + `_moveEnd` as at the end of Leaflet's own zoom animation, plus `updateWhenZooming: false` on the raster layer, back to 12 requests.
5. What would fractional zoom break. Survey of every `getZoom()` consumer: the only hard break is `search-poi`, whose `zoom` parameter is an `Integer` on the server; zoom buttons and zoom-to-fit become fractional, thresholds in the code are all `<`/`>` comparisons.
6. Decision: fractional zoom to hundredths, the server keeps getting an integer. `zoomSnap 0.01`, wheel by raw `deltaY / 450` as MapLibre's `wheelZoomRate` (Leaflet's `getWheelDelta` divides the delta by 3 on Mac, which gave 0.07 per notch instead of 0.22), `Math.round` in `PoiLayer`.
7. Are the comments mentioning MapLibre needed; move the leaflet-hash patch to a util; make `parseMapHash` return one type. Comments trimmed to five one-liners; the patch moved to `map/src/map/util/mapHash.js`, which first broke the map because `"sideEffects": false` dropped an import without exports, fixed by exporting `createMapHash` and `formatMapHash`; the parser returns `null` instead of `false`.

Verified by the agent: lint, jest (413 tests), wheel zoom, zoom buttons and opening the map by a fractional hash in headless Chrome, selenium `map`, `search`, `navigation` and `favorites` suites green; the remaining selenium suites were still running when the PR was opened. Commit and PR made by the agent on the author's request.


alisa911:

RZR-UA:
## AI disclaimer

Produced with Codex (GPT-6).

Prompts used (summarised):
1. Analyze PR #2038 and potential regressions caused by fractional `getZoom()`.
2. Write regression tests, prepare a fix plan, then implement fixes without running tests.
3. Investigate worsening POI jitter on both vector and raster layers, fix it, and add tests compatible with the state before `b69812288646f195f6085c053defab99b7fe8417`.
4. Analyze Chromium-specific MVT flickering with zoom buttons, implement a minimal fix preserving smooth animation, and update regression tests against the previous implementation.
5. Investigate remaining POI jitter during animation and apply a minimal fix without adding a test yet.
6. Analyze tooltip blinking during zoom, propose alternatives, and implement the selected option: preserve the tooltip during animation and reconcile hover afterward.
7. Add regression tests compatible with the state before the tooltip fix.

Decided by the agent, not requested explicitly: reuse the frame-based zoom animation for buttons; use subpixel marker positioning and a fixed geographic zoom anchor; implement shared tooltip hover state with a deferred post-zoom hover check; test tooltip identity and lifecycle using real Leaflet layers with simulated hover events and controlled animation frames.

Tests were not run, as requested. Static lint and formatting checks were performed.

Issue #1867, opened 2026-07-09
Add Close button for context menu open from the list


dmpr0:
## ToDo

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/065d0286-3cc7-4469-86d2-4100c858c027"  width="600" height="auto"/> </td>
  </tr>
</table>   

Currently there is no way to close the context menu that was opened from Search or My Places.

- [x] Add **Close** button
	- [x] Icon – `ic_action_close`
	- [x] Add 20 px padding between Close and Configure Map / Search button group
	- [x] Place before Configure Map
- [x] Show only when the menu opens from Search / My Places, and the main leading action is Back
- [x] Clear section state. For example, clicking Close should clear all Search menu state, so the next click on Search opens a clean search without previous results.
- [x] Add for all types of menus: POI, Favorite, Search, Track...
- [x] Don't move the map back to its previous location after clicking Close.

**Additional:**
- [x] Close menu by clicking the `Esc` button

Dima-1:
<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/01aef28c-5f3a-4523-a3fc-ebaa3fe9d83a"  width="600" height="auto"/> </td>
  </tr>
</table>

PR #2041, opened 2026-09-17
Add close button

Dima-1:
https://github.com/osmandapp/web/issues/1867
## AI disclaimer
Written with Claude Code (Claude Opus 5), driven by [@Dima-1](https://github.com/Dima-1). Summary of the requests it worked from, in order:

Analyze what would be difficult about issue #1867, committing nothing. The agent reported that there is no single close path: the logic sits in closeDetails() in WptDetails.jsx, handleCloseTrackContextMenu() in InformationBlock.jsx, closeHeader() in HeaderHelper.js and backToMainSearch() in SearchResults.jsx, none of which GeneralPanelButtons can reach. It also listed the cases that needed a decision: a track point opened from the Points tab, a wiki/Explore POI, and the local track editor.

Look for functionality that already does something similar, and say which of it fits the issue best. It compared closeHeader(), the ✕ in GroupHeader.jsx and the repeat-click branch of doSelectMenu() in MainMenu.js, and picked doSelectMenu(): it already runs closeSubPages(), honours exitGuards and, unlike the ✕ in GroupHeader, does not call restoreMapView(), which is what the issue asks for. The per-section state reset was taken from that ✕.

Write a plan around doSelectMenu() with minimal changes, then implement it and commit nothing. The plan added a closeMenu flag to AppContext next to the existing closeMapObj, with an effect in MainMenu that clears the section state and calls doSelectMenu(); it also established that ctx.setOpenMenu(item) cannot be used, because updateMenu = !isSelectedMenuItem(item) || ctx.openMenu reopens the section instead of closing it. The implementation touched four files: the flag in AppContext.js, clearMenuState() and the effect in MainMenu.js, the se-close-menu-button and the Esc handler in GeneralPanelButtons.jsx, and a 20px margin in map.module.css.

Cover it with a test. It added the selenium test tests/selenium/src/tests/menu/38-close-menu-button.mjs for a favorite opened from a folder, for Esc, and for a favorite opened from search, then checked the test is real by undoing the fix: without clearMenuState it fails on se-menu-fav-shops, without the search reset on se-default-search-categories.

Show Close on the search results page as well, even when no result is selected. canCloseMenu() got a showInfoBlock parameter and returns early on the results page; a scenario was added to test 38 and confirmed to fail without that line. The file had been edited by hand in the meantime (two comments removed) and the agent kept those edits.

Review the branch, then fix the two bugs the review found. /code-review reported 12 issues, 2 of them bugs. A wiki place could not be reopened after Close because searchSettings.getPoi was never reset and the map layer only opens a place when that value changes; clearMenuState now resets it, and doSelectMenu was switched to a functional setSearchSettings((prev) => ...) so the reset is not overwritten by a stale copy. Clicking a search marker on the map showed Close next to the object's own ✕, where Esc wiped the whole search; the search results branch of canCloseMenu now applies only while the info panel is closed.

Add both cases to test 38, then, on the rest of the review list, "6 is normal by design, fix 7 and 8". The test finds the street marker through window.__leafletMap._layers (as test 29 does) and the wiki place through enumerateIds instead of hard-coded ids, and each case was confirmed to fail with its own fix undone. Item 7 moved the Esc listener and the button into map/src/frame/components/btns/CloseMenuBtn.jsx, dropped the dead ctx.pointContextMenu.element guard and switched canCloseMenu to useSearchNav().isSearchResultRoute; CLOSE_MENU_URLS was left as it is, the agent saying it had overstated that point in its own review. Item 8 removed a misplaced { optional: true } from a waitByRemoved call, added a tracks case and gave the favorites and tracks cases one shared closeMenuKeepingMap helper; the agent also reported that the "map did not move" check still passes with mtx.setMapViewStack([]) undone, so that line is not actually covered by the test.

Review the branch once committed, then: "I checked, Close does not reset the Visible tracks page, the folder list is shown but the URL is /map/visible-tracks/". The review first caught that the branch as committed did not build, because CloseMenuBtn.jsx was untracked while GeneralPanelButtons.jsx imported it, which the amended commit 1ba845f02 closed by adding that one file. The agent had then described the Visible tracks case wrongly, and the correction above was right: the screen and the URL disagree, and it happens without Close too, because TracksMenu decides from its local openVisibleTracks, which resets when MainMenu unmounts the panels, while ctx.openVisibleMenu.showTracks stays true and openVisibleTracksPage() puts /map/visible-tracks/ back. The agent recommended fixing that separately rather than in this PR.

The agent made no commits: it was asked not to, and every commit on the branch, including the one adding CloseMenuBtn.jsx, was made by the author. Test 38, search test 21, eslint and prettier were run by the agent against a local dev server, and behavior in the app was checked both by the agent through screenshots and by the author by hand, whose own check produced the Visible tracks finding. The agent did not open or edit the PR.


map/src/context/AppContext.js
```diff
@@ -345,6 +345,7 @@ export const AppContextProvider = (props) => {
     const [travelShowStartFinish, setTravelShowStartFinish] = useState(false);
 
     const [closeMapObj, setCloseMapObj] = useState(false);
+    const [closeMenu, setCloseMenu] = useState(false);
```
alisa911: is too generic, which menu?

map/src/frame/components/btns/CloseMenuBtn.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+    // close the menu by Esc
+    useEffect(() => {
+        const closeMenuByEsc = (e) => {
+            if (e.key !== 'Escape') return;
+            if (e.defaultPrevented) return;
+            if (e.target.closest?.(ESC_IGNORE_TARGETS)) return;
+            ctx.setCloseMenu(true);
+        };
+        window.addEventListener('keydown', closeMenuByEsc);
```
alisa911: A button should not register a window keydown listener on its own. Move the Esc handling to GeneralPanelButtons

map/src/frame/components/btns/CloseMenuBtn.jsx
```diff
@@ -0,0 +1,33 @@
+import React, { useContext, useEffect } from 'react';
+import { IconButton, Paper, SvgIcon } from '@mui/material';
+import AppContext from '../../../context/AppContext';
+import mapStyles from '../../../map/map.module.css';
+import { ReactComponent as CloseIcon } from '../../../assets/icons/ic_action_close.svg';
+
+const ESC_IGNORE_TARGETS = 'input, textarea, [contenteditable="true"], [role="dialog"], [role="presentation"]';
+
+// Closes the menu with the opened object; render it only while Close is available
+export default function CloseMenuBtn() {
```
alisa911, line 9: Add a tooltip, like FocusToggleBtn

map/src/frame/components/btns/CloseMenuBtn.jsx
```diff
@@ -0,0 +1,33 @@
+import React, { useContext, useEffect } from 'react';
+import { IconButton, Paper, SvgIcon } from '@mui/material';
+import AppContext from '../../../context/AppContext';
+import mapStyles from '../../../map/map.module.css';
+import { ReactComponent as CloseIcon } from '../../../assets/icons/ic_action_close.svg';
+
+const ESC_IGNORE_TARGETS = 'input, textarea, [contenteditable="true"], [role="dialog"], [role="presentation"]';
+
+// Closes the menu with the opened object; render it only while Close is available
```
alisa911: The comment states a rule enforced in GeneralPanelButtons, not here

alisa911 (changes requested):
The approach needs to change. The button visibility is built as a list of urls and states, so every new page under Search or My Places has to be added by hand, and Explore and POI categories are already missing. The requirement is simpler. Close is shown exactly where the menu shows Back. Please derive it from what already decides Back and from the menu routes MainMenu already has, without a separate list in GeneralPanelButtons.

map/src/frame/components/btns/CloseMenuBtn.jsx
Dima-1: https://github.com/osmandapp/web/pull/2041/commits/366ed8eef4dcdf217b0c6e1a954412cb491db4a9

map/src/frame/components/btns/CloseMenuBtn.jsx
Dima-1: https://github.com/osmandapp/web/pull/2041/commits/366ed8eef4dcdf217b0c6e1a954412cb491db4a9

map/src/frame/components/btns/CloseMenuBtn.jsx
Dima-1: https://github.com/osmandapp/web/pull/2041/commits/366ed8eef4dcdf217b0c6e1a954412cb491db4a9

map/src/context/AppContext.js
Dima-1: https://github.com/osmandapp/web/pull/2041/commits/43b7e7d4abfbbe9ba18dde2583d9184e4ef42026

alisa911 (approved):

PR #2043, opened 2026-09-18
Heatmap prototype: review flow, admin export, dev-only header item

alisa911:
Screenshot 2026-09-18 at 13 19 28

The OsmGpx heatmap prototype moves to `prototypes/heatmap.html` (served as `/heatmap.html` as before) with a clearer track review flow, an admin export and a dev-only header item in the map app.

- The page talks to the host it is opened on: the tracks API, the reviews and the sign-in session are those of `location.origin`; the tiles too, falling back to test.osmand.net while a host has none (osmand.net today). `?data=` and `?api=` still override both for a local stand.
- Review tracks: list items are cards with a chevron; the selected track has an accent border in the list, an orange line with a white halo, start and finish dots, the heat at half opacity and a chip at the bottom of the map. A second click on the card or the line, a click on the empty map or Esc drops the selection; a single found track is no longer selected automatically.
- The track card holds the description, "Download GPX" (the original file via `/osmgpx/get-original-file`, named `<track>-<id>.gpx`) and the review form: verdict buttons only choose, "Correct activity" is a list of the real activities from meta and is sent as the `activity` field the server already stores, "Send review" sends verdict, activity and comment, "Clear the verdict" removes it.
- "Sign in" button, shown when the visitor is not signed in, opens the account page of the same host in a new tab; the page re-checks `/me` when it gets the focus back.
- Admin panel bottom right, only when `/me` says admin: a period and the download of the track reviews for it (`reviews.csv.gz?since=`, server side osmandapp/OsmAnd-tools#1726). The tile-loading status moved to the bottom left next to the selected-track chip.
- Review tracks is open and Display is collapsed by default.
- Map app header: `🔥 Heatmap` after Join us, opens `/heatmap.html` in a new tab, only with `develFeatures` (`REACT_APP_DEVEL_FEATURES`).

Checked on a local stand (the page and the map dev server on localhost:3000, the rest proxied to test.osmand.net) on desktop and at phone width. How `prototypes/` gets deployed as `/heatmap.html` is not described in the repos and has to be settled with the deploy.

## AI disclaimer

1. Find how heatmap.html is created and where the tracks come from. Found the hand-written page in `main/static`, the tile builder `heat_build.py` and the OSM GPX downloader `DownloadOsmGPX`; no code changed.
2. Add a "Give us your feedback" button with a dialog, saving the map view and the filters, and answer where the feedback goes and whether anonymous sending is acceptable. Built the dialog and a `POST /api/pubtracks/feedback`; later the author was told to drop the page feedback entirely, so the button, the dialog and the server side were removed again (commit `e42b7f8d8`).
3. Set up a way to test the local page against the test server. A localhost:3000 proxy stand outside the repo.
4. Make the track list look clickable and the selected track as clear as on the map, and solve how a phone user sees the selection. Cards, accent selected state, scroll to the card, a chip over the map; commit `0550991c6`.
5. Add "Download GPX" to the track card; answer original vs simplified line, private tracks and sign-in. Original file through the existing endpoint, no sign-in, private tracks are not in the database. Then make the selected track stand out: halo, start and finish dots, half-opacity heat (dropped once and brought back on request), no automatic selection of a single track. Commit `334f780b7`.
6. A comment was sent only with a verdict click: add a Send button, disabled without a verdict, and let the correct activity be chosen from the existing list and stored as its own field. The page now sends `activity`; the description and the form merged into one card unlike the selected list item. Commit `bf8a25f75`.
7. Display collapsed and Review tracks open by default. Commit `307fea9f2`.
8. Give admins a convenient download of reviews for a period; the author proposed a collapsed bottom-right panel. Built with a period and the download; the tile status moved to the bottom left, the public reviews link left as it was. Commits `2a529b773`, `e42b7f8d8`.
9. Move the page into the web repo under `prototypes/`, keep the URL, make requests and sign-in go to the host the page is opened on, add a header item after Join us that opens in a new tab and only with dev features; replace the "sign in on host" text with a Sign in button opening the account page in a new tab. The agent first edited the docusaurus config and put the item after Pricing; both were corrected on the author's instruction. Commit `c3d0df4a2`.

Commits and pushes were made by the agent on the author's instruction. Verified by the agent on the local stand against test.osmand.net.


alisa911:
https://github.com/osmandapp/OsmAnd-Issues/issues/2708

PR #2050, opened 2026-09-21
Zoom the trackpad at 100 px per level and the mouse at 450 as MapLibre

alisa911:
Since the fractional wheel zoom (7784c4f3c) every `wheel` event was divided by 450 px per zoom level. That is MapLibre's rate for a continuous mouse wheel only; its trackpad rate is 100 px per level, so a laptop trackpad swipe moved the zoom 4.5 times less than it should.

`MapStateLayer.onWheel` now classifies the device as MapLibre's `ScrollZoomHandler` does:

- a `deltaY` multiple of 4.000244140625, or a lone tick after a pause longer than 400 ms, is a mouse: 450 px per level;
- `|deltaY| < 4`, or a repeating event with `|gap * deltaY| < 200`, is a trackpad: 100 px per level;
- the type holds for the gesture, and every event is compressed with MapLibre's sigmoid (at most one level per event).

MapLibre's 40 ms timeout for a lone tick is not ported (project rule: no timers); a lone tick zooms as a mouse right away.

Unit tests in `tests/unit/src/tests/map/01-fractional-zoom.test.js` are updated to the MapLibre formula and cover the classification and a trackpad flick vs. the same wheel distance of a mouse; all four fail with the old handler.

## AI disclaimer

1. Users complain that a trackpad swipe barely moves the zoom since the fractional MVT zoom; look at how Google does it and make it the same. The agent traced the wheel path to `WHEEL_PX_PER_ZOOM = 450` in `MapStateLayer.js`, found that MapLibre uses 450 only for a mouse and 100 for a trackpad, and measured Google Maps with synthetic wheel events in a browser pane (about 200 px per level for small deltas, compressed for large ones). It proposed MapLibre's classification and rates and asked before touching code.
2. Fix it, do not commit yet. `onWheel` got MapLibre's device classification, the two rates and the sigmoid; the 40 ms timeout was replaced by treating a lone tick as a mouse. The unit tests were updated through the add-test skill: two expectations rewritten to the formula, two tests added, all four verified to fail without the fix.
3. Did you test? The agent had only run the unit tests; it then started the dev server and replayed four wheel scenarios in its browser pane: the zoom matched the formula in every case. Starting the dev server opened a tab in the author's Chrome, which the author objected to; the server was restarted with `BROWSER=none`.
4. Where does 4.000244140625 come from? From MapLibre's `scroll_zoom.ts` (`wheelZoomDelta`, the `deltaY` of one mouse wheel tick, 4 + 2^-12); the browser that reports it was not verified.
5. Commit and open the PR.

The commit and this PR were made by the agent on the author's request. The synthetic wheel checks and the unit tests were run by the agent; a real trackpad swipe was not checked by the agent.


alisa911:

